PR #8, opened 2023-08-12
Update interpreter.py - windows file system adjustment

nirvor:
Hi!

I replaced (if/else condition...) the use of the `readline` library with `pyreadline3` bfor Windows compatibility. `readline` is Unix-specific and not working with windows file system, while `pyreadline3` provides similar functionality on Windows.

( Maybe there are better solutions, like checking at building via poetry? Or maybe its fine for now)

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -10,7 +10,13 @@ import platform
 import openai
 import getpass
 import requests
-import readline
+try:
+    if platform.system() == 'Windows':
+        import pyreadline3 as readline
+    else:
+        import readline
+except ImportError:
+    print("Windows file system detected, please pip install pyreadline3.")
 import urllib.parse
 import tokentrim as tt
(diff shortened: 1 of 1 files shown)


nirvor:
Oh, it seems to be fine if you just keep the interpreter.py like it is with "`import readline`". It seems to be fine if the pip install interpreter is installing `pyreadline3` if the OS is any win version. 

Maybe a change in poetry build / pyproject.toml would be better?

```
[tool.poetry.dependencies.pyreadline3]
version = "*"
markers = "sys_platform == 'win32'"

[tool.poetry.dependencies.readline]
version = "*"
markers = "sys_platform != 'win32'"
```

KillianLucas:
Great catch @nirvor! Adding it to the `.toml` now and will keep `interpreter.py` as-is, importing readline. Will push to pip, let me know if it works 👍

PR #25, opened 2023-09-02
Changed model name from 16b to 13b, added tip for gpt3.5

TanmayDoesAI:
Changed the model name from 16b as, the models released by Llama were 13b, probably a typo.
 
Also added a tip to change the model to gpt 3.5, so that people who do not have access to gpt-4 could also be able to use it


KillianLucas:
Thanks @TanmayDoesAI and @Parv3sh! Great catch on the 16B.

To reduce/simplify the text onscreen (since this comes up every time they run `interpreter`), let's just include the `--local` tip for now, unless we can compress both onto one line with just a few words. Maybe the solution is actually to have one tip that very briefly explains how to change the model.

PR #28, opened 2023-09-03
Update README.md

eltociear:
avaliable -> available


KillianLucas:
Thanks @eltociear, good catch!

PR #34, opened 2023-09-04
Change of the OpenAI's tool name

matejkajinic:
OpenAI's Code Interpreter changed the name to "Advanced Data Analysis".

Source:"https://openai.com/blog/introducing-chatgpt-enterprise#:~:text=advanced%20data%20analysis%2C%20previously%20known%20as%20Code%20Interpreter."


KillianLucas:
Thanks @matejkajinic! I think let's wait to make this change. Code Interpreter is well known enough, and especially since it's called Open Interpreter (which might turn out to have been a bad call— but who wants "Open Advanced Data Analytics"??) let's leave the focus on Code Interpreter until the new name becomes the more colloquial one.

PR #90, opened 2023-09-07
Enable the streamlit interface

ReinholdM:
Give a interactive website as a chatbot for open interpreter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ dist/
 
 # Ignore misc directory
 misc/
+
+# Ignore .vscode directory
+.vscode/
--- a/streamlit_app.py
+++ b/streamlit_app.py
@@ -1,0 +1,77 @@
+import interpreter, os
+import streamlit as st
+
+
+def load_api_key():
+        """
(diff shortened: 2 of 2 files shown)


ed1g1tal:
I tested this out and it works well. It gives a very simple, customizable, and convenient alternative to the CLI (which I prefer, for one). The cool thing is, after stating the working directory in the prompt, I was making requests to OI via the UI prompt input to add Streamlit widgets to update the UI in real time. It's shaping up to be as functional, if not more so, than the ChatGPT UI.

ReinholdM:
> 

pls check the interface code in line 36, where I try to use the interpreter function rather than ChatGPT. Thanks again!

ed1g1tal:
@KillianLucas If you want to check it out, this is what I am currently using. It's a Streamlit web app that uses Interpreter and can change the Interpreter settings on the fly. I just run it locally but of course it can be deployed to SL community. I am sure that folks like yourself could vastly improve upon it. 

https://github.com/ed1g1tal/oi-chatbot

Thanks @ReinholdM for the idea!

PR #96, opened 2023-09-07
Fix GPT 3.5 from failing to run commands

Maclean-D:
Currently when trying to run commands with the ```--fast``` flag, GPT3.5 tries to run commands in the console with a exclamation point (!) causing it to fail. To recreate issue run ```interpreter -y --fast``` with a prompt like "create a word document in my downloads folder that says hello world :3".
This pull adds the line ```Never use (!) when running commands.``` to the ```system_message.txt``` which resolves the issue.


KillianLucas:
Hey @Maclean-D, great fix! Have seen `gpt-3.5` make this mistake 5 billion times. Merging now.

PR #124, opened 2023-09-07
Add support for Ollama, Palm, Claude-2, Cohere, Replicate Llama2 CodeLlama (100+LLMs)

ishaan-jaff:
This PR adds support for the above mentioned LLMs using LiteLLM https://github.com/BerriAI/litellm/ 
LiteLLM is a lightweight package to simplify LLM API calls - use any llm as a drop in replacement for gpt-3.5-turbo. 

Example
```python
from litellm import completion

## set ENV variables
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion(model="command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -10,6 +10,8 @@ import time
 import json
 import platform
 import openai
+import litellm
+from litellm import completion
 import getpass
 import requests
 import readline
@@ -307,10 +309,10 @@ class Interpreter:
           time.sleep(2)
           print(Rule(style="white"))
 
-      openai.api_type = "azure"
-      openai.api_base = self.azure_api_base
-      openai.api_version = self.azure_api_version
-      openai.api_key = self.api_key
+      litellm.api_type = "azure"
+      litellm.api_base = self.azure_api_base
+      litellm.api_version = self.azure_api_version
+      litellm.api_key = self.api_key
     else:
       if self.api_key == None:
         if 'OPENAI_API_KEY' in os.environ:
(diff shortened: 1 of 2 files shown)


ishaan-jaff:
As for claude-2, Cohere (LLMs that don't support function callin) we could edit the prompt to include the functions and then parse the output to only fetch the code sample

ishaan-jaff:
@KillianLucas can I get a review on this PR ? 

verified it works locally with gpt-4 and azure 

happy to add docs/tests on this if this initial commit looks good

grexzen:
So sweet. 3 days, new frontend, all the LLMs ever lol. You guys rock.

ygorg:
Thanks for this idea ! I cloned your repo and the `litellm` requirement is missing from `pyproject.toml`. Also I could not use Ollama (I saw it was not implemented).

My two cent (I'm a newcomer in world of github projects and LLMs): I wonder whether in order to support every LLMs, there should be parameters in the CLI like : `interpreter --ollama nameofllamamodel` or `interpreter --azure ...`. Or another thing would be to scan every installed models on the computer but this is another project I guess......

KillianLucas:
Hey @ishaan-jaff! Super excited to get this one fully connected. We're playing with lots more `--model` options and this should make those integrations much easier. Merging now.

@ygorg this is a great idea. in the discord earlier we were riffing on a unified model command:

```
interpreter --model gpt-4
interpreter --model azure-{azure-id}
interpreter --model a_script_that_returns_a_generator.py
interpreter --model http://localhost:8000/ # I think this can point to ollama?
interpreter --model Huggingface/Model
...etc
```

what are your thoughts on something like that?

Issue #66, opened 2023-09-06
Hangs on execution of bash script


Loo0D:
Script hangs (infinite wait) on execution of any shell code commands. Ctrl+C from this point returns to interpreter shell.

Running on Arch Linux, installed with `pipx`

Debug output as follows:

```
Running function:
{
    'role': 'assistant',
    'content': "The user wants to know the filetype of a specific file. We can achieve this by using the `file` command in a shell script. This command will return the type of the 
file.\n\nPlan:\n1. Use the `file` command to determine the type of the file at the given path.\n2. Return the output to the user.\n\nLet's execute the plan.",
    'function_call': <OpenAIObject at 0x7faf5cd4f1d0> JSON: {
  "name": "run_code",
  "arguments": "{\n  \"language\": \"shell\",\n  \"code\": \"file /home/X/z.html\"\n}",
  "parsed_arguments": {
    "language": "shell",
    "code": "file /home/X/z.html"
  }
                                                                                                                                                                                              
  file /home/X/z.html                                                                                                                                                                  
                                                                                                                                                                                              

  Would you like to run this code? (y/n)

  y

Running code:
echo "ACTIVE_LINE:1"
file /home/X/z.html

echo "END_OF_EXECUTION"
```
Script hangs here.

jchia:
#80

jchia:
@Loo0D What is your SHELL environment variable? I had a similar failure running interpreter under fish that I fixed by fixing the shell to bash.

Loo0D:
Confirm fish shell, however I also tried it with bash with same results. Though I did launch that instance of bash from fish so maybe that was a factor. I will experiment further and update.

jchia:
If originally you use fish and then you run bash, the SHELL envvar in bash is still fish. It is the SHELL envvar that matters.

#83

Loo0D:
Confirmed SHELL var sets to bash results in correct behaviour.

endolith:
I fixed this by running `conda env config vars set SHELL=/bin/bash` in my open-interpreter conda environment.  (Please don't assume that everyone runs bash)

PR #136, opened 2023-09-07
Updated README.md to reflect Fish shell issues from #66

Loo0D:


KillianLucas:
Thanks for this early commit @Loo0D! Closing since the README should focus on the core installation / features. Ideally we'd have a help section in the near future with articles that cover much more ground / integrations, and this would be perfect for that.

Issue #88, opened 2023-09-07
Issue in Python Installation


shubhexists:
OS - Ubuntu 20.04
`jerry@jerry:~$ pip install open-interpreter
ERROR: Could not find a version that satisfies the requirement open-interpreter (from versions: none)
ERROR: No matching distribution found for open-interpreter
`

fatelei:
python should be >= 3.10

shubhexists:
Got It Thanks for the help

PR #155, opened 2023-09-08
Added Minimum Python Version Requirement

shubhexists:
Issue #88 
I still didn't find it was mentioned anywhere. Hence I updated it in the Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,11 @@
 
 <br>
 
+⚠️ Required: Python Version >= 3.10
+
 ```shell
 pip install open-interpreter
 ```
-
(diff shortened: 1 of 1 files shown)


KillianLucas:
Thank you @shubhexists! I think this might not deter folks from having this error (people are surprisingly unwilling to read these sorts of things at the top of readmes!)

Luckily the error does give a clear error message (pip should say things like "required python >= 3.10" over a bunch of packages). Closing for now, but I appreciate the contribution to solve this issue.

We will def put this on the documentation site once it's up, probably at docs.openinterpreter.com.

PR #157, opened 2023-09-08
Feature: add cli environment variable

moming2k:
I've implemented a feature that allows the interpreter cli to load configurations either directly from environment variables or from a .env file. 

This enhancement aims to streamline the workflow by setting default values in the shell, eliminating the need to repeatedly input parameters via the CLI.

I'd greatly appreciate any feedback or suggestions you might have regarding these changes.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,14 @@
+# Set to 'True' or 'False' to determine if the code should execute without user confirmation
+INTERPRETER_CLI_AUTO_RUN=False
+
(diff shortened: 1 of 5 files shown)


KillianLucas:
Great work @moming2k. It's super annoying to have to set all this every time! Solves that well.

Soon I think we should explore a config/yaml file, which can include more complex stuff like the system message. `interpreter --config` could open it in a text editor.

jordanbtucker:
Might be a good idea to add a CLI arg to choose a specific `.env` file too.

PR #171, opened 2023-09-08
context window option

Frank40790:
Add the option to modify context window by setting 
- [x] Tested running in python
```python
interpreter.context_window=1048
```
- [x] Tested running in cli
```bash
interpreter --context-window 1048
``` 
or
```bash
interpreter -ctx 1048
```


KillianLucas:
Looks great @Frank40790! This solves a huge problem people are having. Appreciate the testing as well. We might have to do some work with `max_tokens` too— I'll play with that tomorrow.

PR #188, opened 2023-09-08
[WIP] Docker / and overall quality of life improvements

darkacorn:
WIP 

docker build . -t o-i:latest && docker run --gpus=all -it o-i:latest

./scripts/build_llama-cpp-python.sh
 
poetry install

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,13 +19,10 @@ RUN apt-get update && apt-get install -y \
 
 WORKDIR /tmp
 
-RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
-    bash Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda && \
-    rm Miniconda3-latest-Linux-x86_64.sh
+COPY --from=continuumio/miniconda3:23.5.2-0 /opt/conda /opt/conda
 
 ENV PATH="/opt/conda/bin:${PATH}"
 
-ENV MAX_LOCAL_CTX="4096"
 
 RUN conda create -y --name interpreter-default python="$PYTHON_VERSION"
 RUN echo "source activate interpreter-default" > ~/.bashrc
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,9 +28,11 @@ RUN echo "source activate interpreter-default" > ~/.bashrc
 WORKDIR /app
 
 COPY  . /app
+RUN chmod +x /app/scripts/startup.sh
 RUN chmod +x /app/scripts/build_llama-cpp-python.sh
-
-RUN mkdir -p  /root/llama.cpp/models
+RUN mkdir -p /app/logs
+RUN mkdir -p  /root/.local/share/Open_Interpreter/models
+RUN ln -s /root/.local/share/Open_Interpreter/models /app/models
 
 RUN pip install -U pip setuptools
 RUN pip install poetry
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh &
 
 ENV PATH="/opt/conda/bin:${PATH}"
 
-RUN conda create -y --name interpreter-default python="3.11.4"
+RUN conda create -y --name interpreter-default python="$PYTHON_VERSION"
 RUN echo "source activate interpreter-default" > ~/.bashrc
 WORKDIR /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,11 +35,11 @@ COPY  . /app
 RUN chmod +x /app/scripts/startup.sh
 RUN chmod +x /app/scripts/build_llama-cpp-python.sh
 RUN mkdir -p /app/logs
-RUN mkdir -p  /root/.local/share/Open_Interpreter
+RUN mkdir -p /root/.local/share/Open_Interpreter
 RUN mkdir -p /app/models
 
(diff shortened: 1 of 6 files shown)


darkacorn:
https://asciinema.org/a/6JjoYQw6o9FKhnb8vyN0FgIUp
building python llama bindings 
installing poetry 

-----
had to change the basepath for linux compliance w/o a space

https://asciinema.org/a/6JjoYQw6o9FKhnb8vyN0FgIUp

darkacorn:
(p3_10) ➜  interpreter git:(main) docker run -v /home/o_0/aidrive2/opensourcetooling/models:/root/.local/share/Open_Interpreter/models --gpus=all -it o-i:latest

model preservation should work too ..

mrdiamonddirt:
LGTM, tested on windows minor llama-cpp-python problems to be resolved 😊

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    bzip2 \
+    git \
+    ca-certificates \
+    --no-install-recommends && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /tmp
+
+RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
```
noneabove1182, line 22: You can just use the official docker image of miniconda:

COPY --from=continuumio/miniconda3:23.5.2-0 /opt/conda /opt/conda

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    bash Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda && \
+    rm Miniconda3-latest-Linux-x86_64.sh
+
+ENV PATH="/opt/conda/bin:${PATH}"
+
+ENV MAX_LOCAL_CTX="4096"
+
+RUN conda create -y --name interpreter-default python="$PYTHON_VERSION"
+RUN echo "source activate interpreter-default" > ~/.bashrc
+WORKDIR /app
```
noneabove1182, line 32: I suggest, though it may not working, swapping to a cuda-runtime image and then copying conda config over to get a reduced final image size, I use that in my own images for text-generation-webui:

FROM nvidia/cuda:11.8.0-runtime-ubuntu22.04

\# Copy conda and cuda files over
COPY --from=builder /opt/conda /opt/conda
COPY --from=builder /usr/local/cuda-11.8/targets/x86_64-linux/include /usr/local/cuda-11.8/targets/x86_64-linux/include 

ENV PATH=/opt/conda/bin:$PATH

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN mkdir -p /app/logs
+RUN mkdir -p /root/.local/share/Open_Interpreter
+RUN mkdir -p /app/models
+
+#RUN ln -s /root/.local/share/Open_Interpreter/models /app/models
+#RUN ln -s /app/models /root/.local/share/Open_Interpreter/models
+RUN pip install -U pip setuptools
+RUN pip install poetry
+
+CMD [ "bash" ]
```
noneabove1182, line 46: shouldn't this entry run the startup.sh so that it installs llama-cpp-python?

Dockerfile
noneabove1182: I also run the following for maximum compatibility, has worked well for me so far:

`RUN apt-get -y update && apt-get -y install wget && wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/cuda-keyring_1.0-1_all.deb \
    && dpkg -i cuda-keyring_1.0-1_all.deb && apt-get update && apt-get upgrade -y \
    && apt-get -y install python3 build-essential \
    && apt-get -y install cuda-11.8 && apt-get -y install cuda-11.8 \
    && systemctl enable nvidia-persistenced \
    && cp /lib/udev/rules.d/40-vm-hotadd.rules /etc/udev/rules.d \
    && sed -i '/SUBSYSTEM=="memory", ACTION=="add"/d' /etc/udev/rules.d/40-vm-hotadd.rules
`

Dockerfile
```diff
@@ -0,0 +1,46 @@
+ARG CUDA_VERSION="11.8.0"
+ARG CUDNN_VERSION="8"
```
noneabove1182, line 2: have you tested without CUDNN? this is a purely curiousity question, i tried both CUDNN and regular devel and saw no difference for text-generation-webui

Dockerfile
noneabove1182: note that I also had to run 

`RUN apt-get update && apt-get remove --purge -y nvidia-* \
    && apt-get install -y --allow-downgrades nvidia-driver-535/jammy-updates`

because there were some annoying incompatibilities with the nvidia cuda image drivers and the end user drivers

Dockerfile
jordanbtucker: I'm my testing, cuDNN was not required to build llama.cpp.

Dockerfile
darkacorn: correct as its a wip .. just in there for "quick" access

Dockerfile
darkacorn: actually handy - thanks

Dockerfile
darkacorn: i keep it my way so far .. the container is just for maximum compatibility .. 
no need to get N new installs or add udev rules

Dockerfile
darkacorn: cudnn isnt required .. correct - but this is premature optimisation for a temp solution

darkacorn:
closed in fav of #204

PR #189, opened 2023-09-08
Add option to override Azure API type

Taik:
Adding an option to explicitly override `api_type` that's passed to OpenAI when in `--azure` mode. This can be set via `export AZURE_API_TYPE=azure_ad`.

This is needed since our internal Azure instance at Airbnb requires a different "api_type" set.

image


KillianLucas:
Great contribution @Taik. Cool to hear the folks at Airbnb are using OI! Merging.

PR #198, opened 2023-09-08
Distinguishing the built-in print from the imported print

### LINKED ISSUE #130: Distinguishing the built-in "print" from the imported print in interpreter/llama_2.py
In this file: `interpreter/llama_2.py`, `print` has been imported from `rich`: `from rich import print`. How to distinguish it from the built-in `print` function? Isn't there a naming conflict between the two? 

How about renaming the imported print or just use `rich.print`?

markharley12:
This was a small change as a proposed response to resolve  #130

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -15,7 +15,7 @@ import requests
 import readline
 import urllib.parse
 import tokentrim as tt
-from rich import print
+from rich import print as rprint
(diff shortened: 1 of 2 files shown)


ericrallen:
Hey there, @markharley12!

I think this was resolved in the refactor following a similar pattern to what you proposed here.

PR #201, opened 2023-09-08
added a regex content shebang checker that overrides language

### LINKED ISSUE #163: Bash code doesn't run correctly
Sometimes, it writes bash code (e.g. echo test) and tries to run it, and it executes in python and errors out.

markharley12:
It's great to be here. I am a noob junior dev so feel free to roast my PR.

This is a reproducable problem I experienced for bash using the low-7b Code-Llama model.
I noticed it in an issue and I believe it may resolve #163

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -6,18 +6,19 @@ from .code_interpreter import CodeInterpreter
 from .llama_2 import get_llama_2_instance
 
 import os
+import re
 import time
 import json
-import platform
 import openai
 import getpass
+import platform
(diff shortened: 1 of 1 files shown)


KillianLucas:
Hi @markharley12 it's great to have you! I think this particular shebang format isn't usually produced given our prompts (except HTML, which does get produced like that), but I believe there are packages out there that scan code to determine the language. I've had this problem myself with code llama, so your idea is excellent and we should implement it soon.

Great PR and thanks for contributing.

PR #213, opened 2023-09-09
Falcon // Any 🤗 model via `--model meta/llama`

KillianLucas:
Still would really like to refactor into something cleaner, but this does appear to do the trick.

Mimics what we had with CodeLlama— put in `interpreter --model TheBloke/CodeLlama-7B-GGUF`, flip through all the quantization options available for that model, downloads it. Cool!

Works with WizardCoder, Falcon, Llama, Samantha, etc.

Also solved a lot of 'silent download failures' by not using wget or curl, but `huggingface_hub` itself to get the models.


darkacorn:
nice !

PR #220, opened 2023-09-09
Update system_message.txt

moshejs:
"try again and again" is causing an infinite loop without break. this is how I would approach fixing it

### DIFF
--- a/interpreter/system_message.txt
+++ b/interpreter/system_message.txt
@@ -3,7 +3,7 @@ First, write a plan. **Always recap the plan between each code block** (you have
 When you send a message containing code to run_code, it will be executed **on the user's machine**. The user has given you **full and complete permission** to execute any code necessary to complete the task. You have full access to control their computer to help them. Code entered into run_code will be executed **in the users local environment**.
 Only use the function you have been provided with, run_code.
 If you want to send data between programming languages, save the data to a txt or json.
-You can access the internet. Run **any code** to achieve the goal, and if at first you don't succeed, try again and again.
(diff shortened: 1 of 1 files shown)


grexzen:
> Wouldn't it be better to dynamically change the prompt based on comparisons of the output? There should be some kind of system that recognizes the result of the command output (error code) is identical for multiple runs.

Great idea! Error handling is now dynamic prompting.

interpreter/system_message.txt
```diff
@@ -3,11 +3,11 @@ First, write a plan. **Always recap the plan between each code block** (you have
 When you send a message containing code to run_code, it will be executed **on the user's machine**. The user has given you **full and complete permission** to execute any code necessary to complete the task. You have full access to control their computer to help them. Code entered into run_code will be executed **in the users local environment**.
 Only use the function you have been provided with, run_code.
 If you want to send data between programming languages, save the data to a txt or json.
-You can access the internet. Run **any code** to achieve the goal, and if at first you don't succeed, try again and again.
+You can access the internet. Run **any code** to achieve the goal, and if at first you don't succeed, try again and again but by the 3rd iteration try something else.
```
ericrallen, line 6: Hey there, @moshejs!

If you are still noticing this issue in the most recent version, the system message is now handled in [`interpreter/config.yaml`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/config.yaml#L1) after the codebase was refactored.

moshejs: thank you, will let you know

PR #230, opened 2023-09-10
remove redundant checks for apple silicon

shubhe25p:
platform.machine() will check if its apple silicon as well as if python is compatible. Removed the extra check and fixed some comments.


KillianLucas:
Looks great! Many thanks @shubhe25p, merging now.

PR #235, opened 2023-09-10
Update contributing.md with instructions on how to get local fork running

oliverpalonkorp:
Added quick instructions to get a local fork running with poetry. It currently is not obvious to non-poetry users that this is how you should run your local fork.


KillianLucas (approved):
Great work @oliverpalonkorp. Struggled with this myself!

PR #265, opened 2023-09-11
feat(interpreter): added error message for response

codeacme17:
Add user prompts for openai's response errors and exit in non-debug mode
Current implementation: `AuthenticationError` and `Timeout`

- AuthenticationError
  scree-shot 2023-09-11 162526

- Timeout
  scree-shot 2023-09-11 162312

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -27,6 +27,7 @@ from .get_hf_llm import get_hf_llm
 
 
 import os
+import sys
 import time
 import traceback
 import json
@@ -263,8 +264,12 @@ class Interpreter:
     if message:
       # If it was, we respond non-interactivley
       self.messages.append({"role": "user", "content": message})
-      self.respond()
-
+      
+      try:
+        self.respond()
+      except Exception as e:
+        self._print_error_message_and_exit(e)
+      
     else:
       # If it wasn't, we start an interactive chat
       while True:
@@ -479,7 +484,6 @@ class Interpreter:
       
       for _ in range(3):  # 3 retries
         try:
-
             if self.use_azure:
               response = litellm.completion(
                   f"azure/{self.azure_deployment_name}",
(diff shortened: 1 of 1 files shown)


interpreter/interpreter.py
```diff
@@ -834,4 +838,15 @@ def messages_to_prompt(messages):
 
   def _print_welcome_message(self):
     current_version = pkg_resources.get_distribution("open-interpreter").version
-    print(f"\n Hello, Welcome to [bold white]⬤ Open Interpreter[/bold white]. (v{current_version})\n")
\ No newline at end of file
+    print(f"\n Hello, Welcome to [bold white]⬤ Open Interpreter[/bold white]. (v{current_version})\n")
```
codeacme17, line 841: When processing this line, a conflict will occur. Please select the code of the main branch.

jordanbtucker:
Thanks for this. Will it still print the stack trace? That information is critical at this point in development. Once we get some proper logging it won't be as critical.

codeacme17:
@jordanbtucker, You're right, I've reserved stack tracing for debug mode. Currently, the error will only occur when there's a response error from llm. In non-debug modes, stack traces will still appear for other errors.

jordanbtucker:
Awesome. I think for specific errors like `openai.error.AuthenticationError` and `openai.error.Timeout`, I think it's fine to leave the stack traces out. But especially when catching generic Exceptions we should include the stack traces.

Also, probably not a good idea to call our users dummies for entering an invalid API key.

image

(I'm joking, of course 😁)

codeacme17:
😁 If possible, could you please take a look at PR #312? I think it can be merged.

ericrallen:
Hey there!

I'm not entirely sure this change is still valid and it would need to be updated to account for the codebase refactor, so I'm going to close it as stale for now, but feel encouraged to reopen it if you feel it's still worth pursuing and want to update it for the new codebase structure.

PR #273, opened 2023-09-11
Remove previous message and its responses from chat history with undo-command.

oliverpalonkorp:
Here's my implementation of removing previous messages by undo -command.

The current implementation works the following way.
The user writes `%undo`, which calls `remove_previous_conversation` -function. This function finds the latest `role: user` message, and removes everything that was added to `messages` after that (and the user's prompt too). This way the user doesn't have to remove individual messages, responses and code blocks (which can be confusing for the context), but can rather just undo their last prompt – similar to what a user would do in ChatGPT's UI when editing a previous message: nuking everything after that.

The reason I implemented this with the `%undo` command, is that hotkeys are a bit of a grayzone for the project right now. I have indeed added a `hotkeys.py` HotkeyHandler, which works, but there are some issues:

1. The hotkeys need user premission to be listened to from System Settings – which I was not sure is in the interest of the project, as it adds some friction for users wanting to use OI.
2. I have not tested the hotkey implementation on Windows, as I do not currently have access to a Windows machine (feel free to test).

Rows 303-307 in `intepreter.py` contain comments and code to activate the hotkey -listener for CMD+Z. However, this should not be activated until a hotkey for Windows is made, and a distinction between them. We can't have a universal CTRL+Z, as on Mac it kills the Terminal process by default, and cannot be overwritten afaik.

We can revisit and build on the hotkey functionality, but for now I would advice only to merge the `remove_previous_conversation` function and the check for the `%undo` command on rows 322-325 of `interpreter.py`.

The hotkey functionality also adds a new dependency `pynput`, which is not needed if hotkeys are not activated.

**UPDATE:** Removed hotkey functionality. This can be revisited later, but for now, this should go well with @moming2k 's command interface.

- Oliver


oliverpalonkorp:
Now the test seems to fail because of a lack of API key. Shouldn't be an issue with my code though?

moming2k:
@oliverpalonkorp I've made some refactoring to the command handle in [PR #302](https://github.com/KillianLucas/open-interpreter/pull/302). I've ensured it doesn't impact the functionality you've written here. If your code gets merged first, I'll gladly integrate the new command handle accordingly. Thank you for your understanding!

oliverpalonkorp:
Fixed. Should be ready to merge.

interpreter/interpreter.py
```diff
@@ -189,13 +188,46 @@ def reset(self):
...
+
   def handle_help(self, arguments):
     commands_description = {
       "%debug [true/false]": "Toggle debug mode. Without arguments or with 'true', it enters debug mode. With 'false', it exits debug mode.",
       "%reset": "Resets the current session.",
       "%save_message [path]": "Saves messages to a specified JSON path. If no path is provided, it defaults to 'messages.json'.",
       "%load_message [path]": "Loads messages from a specified JSON path. If no path is provided, it defaults to 'messages.json'.",
-      "%help": "Show this help message."
+      "%help": "Show this help message.",
+      "%undo": "Remove previous messages and its response from the message history.",
```
moming2k, line 230: Maybe relocate it to below %reset, as they are the same type of command, and leave the help as the last item

README.md
```diff
@@ -227,15 +227,16 @@ $ interpreter
...
+ • `%debug [true/false]`: Toggle debug mode. Without arguments or with 'true', it
+enters debug mode. With 'false', it exits debug mode.  
+ • `%reset`: Resets the current session.  
+ • `%save_message [path]`: Saves messages to a specified JSON path. If no path is
+provided, it defaults to 'messages.json'.  
  • `%load_message [path]`: Loads messages from a specified JSON path. If no path  
-   is provided, it defaults to 'messages.json'.                                 
+ is provided, it defaults to 'messages.json'.  
  • `%help`: Show the help message.
+• `%undo`: Remove previous messages and its response from the message history.
```
moming2k, line 239: Maybe relocate it to below %reset, as they are the same type of command, and leave the help as the last item

interpreter/interpreter.py
```diff
@@ -189,13 +188,46 @@ def reset(self):
...
+        self.messages = self.messages[:last_user_index]
+
+    print("") # Aesthetics.
+
+    # Print out a preview of what messages were removed.
+    for message in removed_messages:
+      if 'content' in message and message['content'] != None:
+        print(Markdown(f"**Removed message:** `\"{message['content'][:30]}...\"`"))
+      elif 'function_call' in message:
+        print(Markdown(f"**Removed codeblock**")) # Could add preview of code removed here.
```
moming2k, line 219:
```suggestion
        print(Markdown(f"**Removed codeblock**")) # TODO: Could add preview of code removed here.
```

oliverpalonkorp:
Thanks @moming2k . Committed your changes.

KillianLucas:
Great work @oliverpalonkorp. Brilliantly done-- I could use this a lot myself. Merging now.

PR #288, opened 2023-09-11
[Litellm] Add claude 2 + Together AI

ishaan-jaff:
https://github.com/KillianLucas/open-interpreter/assets/29436595/4268181c-0b39-4543-a855-1cbf0d98f3e0

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,12 +150,17 @@ Run shell commands with -y so the user doesn't have to confirm them.
 print(interpreter.system_message)
 ```
 
-### Change the Model
-
+### Change the Model - We support 100+ LLMs
+open-interpreter uses [LiteLLM](https://github.com/BerriAI/litellm) for LLM Input/Ouput
 For `gpt-3.5-turbo`, use fast mode:
 
 ```shell
 interpreter --fast
+```
+
+For `claude-2`, use
+```shell
+interpreter --model claude-2
 ```
 
 In Python, you will need to set the model manually:
--- a/interpreter/cli.py
+++ b/interpreter/cli.py
@@ -174,6 +174,9 @@ def cli(interpreter):
     if "/" in interpreter.model:
       interpreter.local = True
 
+    if "togethercomputer/" in interpreter.model:
+      interpreter.local = False
+
   if args.api_base:
     interpreter.api_base = args.api_base
 
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -43,7 +43,7 @@ from rich import print
 from rich.markdown import Markdown
 from rich.rule import Rule
 
-# Function schema for gpt-4 
+# Function schema for gpt-4
 function_schema = {
   "name": "run_code",
   "description":
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -489,7 +489,6 @@ class Interpreter:
       
       for _ in range(3):  # 3 retries
         try:
-
             if self.use_azure:
               response = litellm.completion(
                   f"azure/{self.azure_deployment_name}",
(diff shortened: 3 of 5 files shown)


ishaan-jaff:
cc @KillianLucas

README.md
```diff
@@ -150,14 +150,19 @@ Run shell commands with -y so the user doesn't have to confirm them.
...
 
-### Change the Model
-
+### Change the Model - We support 100+ LLMs
+open-interpreter uses [LiteLLM](https://github.com/BerriAI/litellm) for LLM Input/Ouput
 For `gpt-3.5-turbo`, use fast mode:
 
 ```shell
 interpreter --fast
 ```
```
ishaan-jaff, line 159: can we standardize model selection for gpt-3.5 to be 

```
interpreter --model gpt-3.5-turbo
```

jordanbtucker: Agreed. `--fast` should be deprecated.

But I don't particularly like `--model` being used for API access. I'd prefer `--local --model` to be used for local models and something like `--api` to be used to specify an API like OpenAI GPT-4 or Claude-2. It makes it more clear whether you're sending your data to the cloud or not.

Just my two cents.

ishaan-jaff: i agree i like --api too !

KillianLucas: i think `api` is less clear than `model`, i think choosing your makes sense! i love `--model claude-2`.

agreed it's important to make the distinction very clear in our messaging though. `--local` should always be on for local models, if off, we're on the cloud (which is the way OI is intended to be used / default behavior).

and agreed `--fast` should probably be deprecated (unless we decide it's a nice shortcut to quickly use gpt-3.5 (i use it this way) + can think of other shortcuts that are similar, so this isn't a weird one-off.)

(local _was_ like this in that it autofilled model as code-llama, but now i think it's better as a bool re: we'll attempt to run this model locally vs in the cloud)

codeacme17: How about giving users a way to choose? When the user directly enters `interpreter`, a list of models is listed for the user to choose.

codeacme17: I completely understand where you're coming from, and I genuinely agree with your perspective. And, I feel that having a list might ease the mental load for our users and simplify our code, essentially reducing the entry points for selecting a model. Your method seems quite feasible, and if those configurations remain confidential, I believe there shouldn't be any issues. Just sharing my thoughts.

RichardAwoyemi: > and agreed `--fast` should probably be deprecated (unless we decide it's a nice shortcut to quickly use gpt-3.5 (i use it this way) + can think of other shortcuts that are similar, so this isn't a weird one-off.)

Fwiw, I also use --fast because it's simple, less friction and makes things that bit faster.

ishaan-jaff: will address that in this PR:

- [ ] interpreter --models shows a list of supported llms that a user can select 
- [ ] If a user has entered keys `interpreter --models` should show llms supported for their validated keys (no security issues)
- [ ] If a user has not entered any keys, `interpreter --models` should show 5 llms from popular providers - `anthropic/claude-2`, `anthropic/claude-1.2`, `cohere/command-nightly`, `togetherai/llama2`, `ai21/j2-ultra`

ishaan-jaff: @KillianLucas open questions:
- What should go in the config file ? 
- Should I address parsing the code generated by claude-2 in this PR ? (similar to how you're handling local llms)

krrishdholakia:
Can you enable custom api bases too @ishaan-jaff 

```
interpreter run --model claude-2 --api_base `my_custom_api_base`
```

haseeb-heaven:
I am getting error even after running Claude 2 and Vertex AI Code Bison.

```cpp
haseeb-mir@Haseebs-MacBook-Pro open-interpreter % python3.11 -m poetry run interpreter --model code-bison
A new version is available. Please run 'pip install --upgrade open-interpreter'.

▌ Model set to CODE-BISON                                                                                                                                             

Tip: To run locally, use interpreter --local                                                                                                                            

Open Interpreter will require approval before running code. Use interpreter -y to bypass this.                                                                          

Press CTRL-C to exit.                                                                                                                                                   

> Write factorial of number in python and save it to file called fact_num.py and then try to run the program.
  File "<string>", line 1, in <module>
  File "/Users/haseeb-mir/Documents/Code/Python/open-interpreter/interpreter/interpreter.py", line 131, in cli
    cli(self)
  File "/Users/haseeb-mir/Documents/Code/Python/open-interpreter/interpreter/cli.py", line 207, in cli
    interpreter.chat()
  File "/Users/haseeb-mir/Documents/Code/Python/open-interpreter/interpreter/interpreter.py", line 412, in chat
    self.respond()
  File "/Users/haseeb-mir/Documents/Code/Python/open-interpreter/interpreter/interpreter.py", line 584, in respond
    messages = tt.trim(self.messages, self.model, system_message=system_message)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/haseeb-mir/Library/Caches/pypoetry/virtualenvs/open-interpreter-hpQxQ_Bl-py3.11/lib/python3.11/site-packages/tokentrim/tokentrim.py", line 133, in trim
    raise ValueError(f"Invalid model: {model}. Specify max_tokens instead")
ValueError: Invalid model: code-bison. Specify max_tokens instead
```

kingpower2542:
import os 

os.environ["OPENAI_API_KEY"] = "your-api-key"

ericrallen:
I'm going to close this one as I think it's no longer relevant after the refactor and recent updates to the integration with LiteLLM.

Please feel encouraged to reopen this Pull Request if I'm wrong.

PR #292, opened 2023-09-12
Fix typo.

Michael-Lfx:


KillianLucas (approved):
ah, right. thanks for catching this @Michael-Lfx!

PR #304, opened 2023-09-12
Added %reset -> self.reset() and %exit -> exit(0) to interactive chat…

CodeAKrome:
… loop


jordanbtucker:
@CodeAKrome Any reason why you closed this? This seems like a great addition. I think we already have reset, but we don't have exit.

CodeAKrome:
I thought it was a dupe of another PR. I'll add %exit back.

PR #305, opened 2023-09-12
Add Dockerfile and instructions to run with docker

MarcoWorms:
Maybe this helps easily contain any dependency issues when installing, and also sandboxes the interpreter so it can interact with our machines more safely!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM python:3.10
+
+WORKDIR /app
+
+ADD . /app
+
+RUN pip install poetry
+
+RUN poetry install
+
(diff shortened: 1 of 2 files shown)


moming2k:
Interest to know why it need to expose 80 port ?

jordanbtucker:
Thanks for the suggestion. There is already docker support in the works in #204

MarcoWorms:
> Interest to know why it need to expose 80 port ?

maybe not needed, I ported the Dockerfile from another place so is probably a leftover :P

Nice that you are already working on supporting it, will close this pr then :) thanks!

PR #307, opened 2023-09-12
added ci workflow for linting and formatted the code with black

0xcha05:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          python-version: "3.11"
+
+      - name: Install Black
+        run: |
+          python -m pip install --upgrade pip
+          pip install click==7.1.2
+          pip install black==21.4b2
+
+      - name: Run Black
```
astelmach01, line 27: Personally, I would avoid actually modifying code in the CI. Instead, flake8 should be used as a linter. 

I would go with pre-commit hooks instead to ensure committed code is formatted.

[My pre commit hooks](https://github.com/astelmach01/Cosmo-backend/blob/main/.pre-commit-config.yaml) is more than enough.

astelmach01 (changes requested):

.github/workflows/lint.yml
0xcha05: it doesn't modify the code, just does a check.

astelmach01: Ah you're right, I would still add pre commit hooks nonetheless

gowtham-ryze: This is widely used and a common practice. if you want to have both this and pre-commit hooks, we can add that in a separate PR.

0xcha05: it's a pretty common practice, if you want both this and pre-commit hooks, we can do it in a different PR.

PR #309, opened 2023-09-13
feat: add automated changelog with release please

eden-chan:
Might be a good idea to set up `staging`

generate release PRs based on the conventionalcommits.org spec
https://github.com/googleapis/release-please

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,19 @@
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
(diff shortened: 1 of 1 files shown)


jordanbtucker:
Thanks for the suggestion. I think this will be a great idea once this project settles in a little more. For now, I think we want to keep a more hands-on approach with our changelogs.

KillianLucas:
As @jordanbtucker mentioned, best to have more control over change-log at this stage. Thanks for the contribution @eden-chan! I'm sure we'll do something like this in the future.

PR #311, opened 2023-09-13
Added support for LM Studio to Open Interpreter

PaulConyngham:
This means Open Interpreter can now use any Open Source LLM off Huggingface produced by folks such as @TheBloke

video of it in action here:

https://twitter.com/paul_conyngham/status/1701755591562326097


PaulConyngham:
https://github.com/KillianLucas/open-interpreter/assets/10569375/4ffbf1db-ad60-429a-bd82-62be716b536a

PaulConyngham:
You can get LM Studio here:

https://lmstudio.ai/

I am not associated with the project.

PaulConyngham:
run this pull request with the following on the command line:

interpreter --lmstudio

mogelbaum89 (approved):

PaulConyngham (commented):
this looks good

jordanbtucker:
Thanks for this PR. Can you describe more about what it does and what the use case is?

Open Interpreter already uses LiteLLM, which supports a huge library of LLMs and APIs including Anthropic, Huggingface, Cohere, TogetherAI, Azure, OpenAI.

PaulConyngham:
Hi Jordan. Did you check what[ LM studio is](https://lmstudio.ai/)? With this PR Open Interpreter can now use any Open Source LLM off Huggingface including model architectures outside of Llama - including Falcon, MPT, Starcoder, Replit, GPT-neoX

PaulConyngham:
described another way, LM studio allows the hosting of local LLMs to be abstracted to a seperate system. By importing & heavily leaning on LiteLLM you are importing significant complexity into the Open Interpreter code base.

Solsboer:
Hi, i would like to see this integration, because llm studio is the only way I could run my LLM using AMD GPU (via OpenCL). If the interpreter could get proper using amd GPU it would be great, but If not then this integration is a very good opportunity. 
I have concerns about this PR - overwriting openai integration seems like not the best way to do this. Probably if it will be done as a real separate option i hope this PR could be merged, for how this is looks like a dirty way of doing this.

PaulConyngham:
hi @Solsboer. I think the opposite. This pr using a -- flag to switch on LMstudio. OpenAI integration is still available without the flag. I think this is actually a much much cleaner way of doing it, because here we are only using the Openai python package and not having to rely on some external system to make calls to local LLMs ala liteLLM.

If you go through and read the code base, about 80% of the code is dedicated to making OI work through Litellm.

Solsboer:
@PaulConyngham I agree that relying on an external system would be nice to avoid, but not sure if it is easy to achieve. During my tests, I wasn't able to send a second message, because according to logs OI sends all chat history and the first LM response with a role "assistantassistantassistantassistant.." which is crashing LMstudio. Also simple prompt won't work correctly even to "hello" message (you can see it in your video example as well), the only way I get is to send 
"### Instruction: Hello.\n###Response:", which shouldn't be a normal prompt. 
I tried to use WizardCoder, Platopys2-13b, and CodeLama. Please let me know if you have different results and if you can drop a video with it, thanks.

smanojreddy:
Yes, I have the same problem, after the first message LM Studio crashes every time, doesn't matter which model I use.

PaulConyngham:
working on it @Solsboer & @smanojreddy. Will post something here shortly

RandomInternetPreson:
LM studio does not support gptq models though? I just downloaded it and it specifically says gguf.

sgaseretto:
Is this functionality going to be added to open-interpreter?

PaulConyngham:
@sgaseretto Hoping so! Have been pestering the author of LM Studio , "Yags", for some time now to support and implement the OpenAi functions capability of the OpenAi python package - he says he is working on it.

jackfood:
any support for LM studio as of current?

ericrallen:
Hey there, folks!

Since this PR for using LM Studio seems like mostly just a change to the `api_base` does that mean that it’s available via the underlying LiteLLM integration?

Nivek92:
@ericrallen LM Studio provides an OpenAI-like API, so LiteLLM should work with it.

ericrallen:
Hey there, folks!

Now that LM Studio is the underlying local model provider, I _think_ we can close this one.

Please reopen it if I'm wrong.

Issue #303, opened 2023-09-12
The code it generated repeated the same function call 20-30 times and used up $5 in a matter of minutes


nayan924:
### Describe the bug

same line was written multiple times, I was pressing y and enter and didnt realise i executed a code with 20-30 calls to a gpt-4 summary of long text and it ate $5.

### Reproduce

I am not sure, but i did paste a code snippet along with instruction to guide it because it was making same mistake multiple times.

### Expected behavior

i wanted to recreate this https://platform.openai.com/docs/tutorials/meeting-minutes/creating-an-automated-meeting-minutes-generator-with-whisper-and-gpt-4

### Open Interpreter version

0.1.3

### Python version

3.11.4

### Operating System name and version

mac 12.5

transcendr:
Not sure about your specific issue but just a tip, instead of pasting stuff to, just tell it to run `pbpaste` and get the text from your clipboard (or whatever the windows equivalent is if on windows)

grexzen:
Where do you set the token limit per call?

Videoteq:
> Where do you set the token limit per call?

Good ideas, but ideally OpenAI should have an enhanced useage limit settings per API key - so that one could set a limit on each API key generated. This would take the heat off OpenInterpreter and also would benefit all OpenAI users. Right now their udeage limits are pretty basic. Using interpreter I clocked up a couple of $ in seconds as it was so exiting! Now I use --fast to use the *much* cheaper GPT3.5-Turbo.

jouky2:
umm, why are you paying?
[**https://github.com/xtekky/gpt4free**](https://github.com/xtekky/gpt4free)

tbels89:
> umm, why are you paying? [**https://github.com/xtekky/gpt4free**](https://github.com/xtekky/gpt4free)

have you implemented this as a local model for use?

ericrallen:
Hey there, @nayan924!

Can you clarify what you’re asking for from Open Interpeter here or can we close this Issue?

ericrallen:
Going to close this one as stale.

Please feel encouraged to reopen it if you’d like to explain the ask here a bit and when we dig into it.

PR #316, opened 2023-09-13
add budget manager for api calls

krrishdholakia:
Addressing this issue https://github.com/KillianLucas/open-interpreter/issues/303

using LiteLLM's budget manager - https://docs.litellm.ai/docs/budget_manager, this PR creates a budget for a session (default set to $10). If this amount gets exceeded, it raises an Exception. 

Confirming it works locally: 
Screenshot 2023-09-12 at 10 21 55 PM


WilliamEspegren:
Neat feature, well needed 👍

krrishdholakia:
cc: @KillianLucas @jordanbtucker let me know if there's anything further required on my end

jordanbtucker:
@krrishdholakia This is awesome! How does a user set their budget?

krrishdholakia:
I didn't expose this as a user controllable setting in the initial PR, but that's a great idea. I'll make updates. 

Re: switch to local - yup sounds good, so check local -> if a model exists -> switch to that -> else raise exception?

jordanbtucker:
I wouldn't even worry about checking if a model exists. Just take them to the local menu after explaining that they've reached their configured budget. And maybe let them know how to change that budget.

Lastly, I think we should remove the default budget and make this an opt-in feature. No budget by default, but they can configure one if they want.

krrishdholakia:
Thanks for the feedback, let me make some changes and update this PR @jordanbtucker

krrishdholakia:
Hey @jordanbtucker, 

Simplified the code based on yours + other feedback. 

Budget manager is now just: 
```
litellm.max_budget = max_budget # your max budget
```

You can set it by doing: 
```
interpreter --max_budget 0.001 # sets a budget of $0.01 
```

It raises a custom error: `litellm.BudgetExceededError`, that i have a try-except around. 

If true, i set `self.local=True`.

krrishdholakia:
Need advice on the last part, as that throws an error due to the menu workflow being triggered at the start of the code interpreter process (before i pass in my openai api key). 

Is there a way for me to trigger this menu easily or do i need to refactor the code and make that into a separate function.

jordanbtucker:
This is looking great! Just so I understand, is this a session-based budget? Like it resets every time the user restarts OI, or does it keep track between OI runs?

One more thing I thought of, it would be nice if users could set their default budget using an environment variable which can be overwritten by the `--budget` arg. Sorry for all the change requests.



> Is there a way for me to trigger this menu easily or do i need to refactor the code and make that into a separate function?

Probably needs to be refactored, but I can take a closer look if you need.

krrishdholakia:
It's a session-based budget -> the module gets set with it when it's initialized. 

re: feedback - 
```
os.environ["LITELLM_MAX_BUDGET"] = 100 # sets the budget to $100
```

Would this be in line with what you were thinking?

jordanbtucker:
That works! Eventually I plan to get away from environment variables for casual users and go with a config file and a CLI for managing the configuration.

krrishdholakia:
oh - say more? @jordanbtucker what's the hypothetical code look like?

krrishdholakia:
@jordanbtucker updated based on requests. 

Menu now works. 

It also checks os.getenv for LITELLM_MAX_BUDGET

KillianLucas:
Incredible work @krrishdholakia. Lots of people's biggest issue with Open Interpreter is token cost. Excited to really highlight this feature soon and thank you + LiteLLM. Merging now.

Issue #321, opened 2023-09-13
Download starts without checking the free disk space


astares:
### Describe the bug

If you run

```
$ pip install open-interpreter
$ interpreter
```
and request the download of the large Llama model the disk size is not checked
(I tried on an old Linux installation which had only 12 GB left, but model size is around 34 GB)

### Reproduce

- use a Linux system that has not much disk space left, one can check with "df -h")
- instruct open-interpreter to use the large parameter model (34B model) with 35.9 GB download

### Expected behavior

- check the remaining disk size **before** the download and warn the user

### Screenshots

-

### Open Interpreter version

0.1.3

### Python version

3.11.4

### Operating System name and version

Linux

### Additional context

-

michaelzdrav:
[This](https://github.com/KillianLucas/open-interpreter/pull/323) should fix the issue. Feel free to clone and test locally with the steps in https://github.com/KillianLucas/open-interpreter/blob/main/docs/contributing.md

Worked for me.

jordanbtucker:
Fix merged and will be in next release!

PR #323, opened 2023-09-13
Check disk space before downloading models

michaelzdrav:
Hi all,

As mentioned in [Issue 321](https://github.com/KillianLucas/open-interpreter/issues/321), there is a bug when downloading models. There is no check of free disk space on the current disk prior to downloading, which can be problematic.

I have introduced this as a potential fix, allowing disk space to be compared to the model size in GB.

Feel free to amend!


KillianLucas:
Oh yeah! Of course we should be doing this. Looks great @michaelzdrav, merging now.

PR #335, opened 2023-09-13
Update GPU.md

metantonio:
Added Integration with Visual Studio, without that CMake won't recognize CUDA Tools, and installation of llama-cpp-python will be incomplete


metantonio (commented):
Integration with Visual Studio, is a feature that must be installed, that enables CMakes to recognize CUDA tools.

PR #336, opened 2023-09-13
Feature voice

mrdiamonddirt:
use --voice to use voice, not sure if will need to the requirements adding to the dependences,
would love for someone to review and point out any silly mistakes i might be making, changes that i need to make

Wakeup word is interpreter

This feature means alot to me as i have family that are blind/partially sighted and would love if this made technology more accessible for anyone

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -346,12 +346,11 @@ class Interpreter:
     if welcome_message != "":
       if welcome_message.startswith(">"):
         print(Markdown(welcome_message), '')
-        if self.voice:
-          assistant.tts_and_play_audio("welcome to open interpreter")
       else:
         print('', Markdown(welcome_message), '')
-        if self.voice:
-          assistant.tts_and_play_audio("welcome to open interpreter")
+        
+      if self.voice:
+        assistant.tts_and_play_audio("welcome to open interpreter")
 
     # Check if `message` was passed in by user
     if message:
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -43,6 +43,9 @@ from rich import print
 from rich.markdown import Markdown
 from rich.rule import Rule
 
+# create an instance of the speech assistant
+assistant = SpeechAssistant(wakeup_word="interpreter")
+
 # Function schema for gpt-4
 function_schema = {
   "name": "run_code",
@@ -108,6 +111,7 @@ class Interpreter:
     self.azure_api_version = None
     self.azure_deployment_name = None
     self.azure_api_type = "azure"
+    self.voice = False
 
     # Get default system message
     here = os.path.abspath(os.path.dirname(__file__))
@@ -342,8 +346,12 @@ class Interpreter:
     if welcome_message != "":
       if welcome_message.startswith(">"):
         print(Markdown(welcome_message), '')
+        if self.voice:
+          assistant.tts_and_play_audio("welcome to open interpreter")
       else:
         print('', Markdown(welcome_message), '')
+        if self.voice:
+          assistant.tts_and_play_audio("welcome to open interpreter")
 
     # Check if `message` was passed in by user
     if message:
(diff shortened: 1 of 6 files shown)


interpreter/interpreter.py
```diff
@@ -342,8 +346,12 @@ def chat(self, message=None, return_messages=False):
     if welcome_message != "":
       if welcome_message.startswith(">"):
         print(Markdown(welcome_message), '')
+        if self.voice:
+          assistant.tts_and_play_audio("welcome to open interpreter")
       else:
         print('', Markdown(welcome_message), '')
+        if self.voice:
+          assistant.tts_and_play_audio("welcome to open interpreter")
```
moming2k, line 354:
```suggestion
    if welcome_message != "":
      if welcome_message.startswith(">"):
        print(Markdown(welcome_message), '')
      else:
        print('', Markdown(welcome_message), '')
        
      if self.voice:
        assistant.tts_and_play_audio("welcome to open interpreter")
```

moming2k:
Please check if the logic is correct and still as expected after change it into one call ?

interpreter/interpreter.py
```diff
@@ -43,6 +43,9 @@
 from rich.markdown import Markdown
 from rich.rule import Rule
 
+# create an instance of the speech assistant
+assistant = SpeechAssistant(wakeup_word="interpreter")
```
moming2k, line 47: would this one still create a listener to the Mic if there is not enable voice?

mrdiamonddirt: yeah sorry it was there abilteraly i will put it somewhere more usefull no jokes 😁

mrdiamonddirt (commented):
was still not a good idea to put this where i did, it's better now

interpreter/voice.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def start_speech_recognition(self):
+        while True:
+             print("Listening for wakeup word...")
+             with SR.Microphone() as mic:
+                   try:
+                        self.speech_recognizer.adjust_for_ambient_noise(
+                            mic, duration=0.4)
+                        audio = self.speech_recognizer.listen(mic)
+                        text = self.speech_recognizer.recognize_google(
```
D0han, line 41: This seems to depend on online service (similar to text to speech above?). That means it leaks information even if open-interpreter is used with local LLM.

haseeb-heaven: There is local TTS Services as well but changes from system to system so we should use that instead

mrdiamonddirt: What do we think of pyttsx3 looks like a good alternative to gtts can run locally and should work well on windows, Mac and linux?

haseeb-heaven: > What do we think of pyttsx3 looks like a good alternative to gtts can run locally and should work well on windows, Mac and linux?

I have used on mac we need to have a wrapper library which check which OS and which package to use from python because sometimes it doesnt speak no voice available i had issue with my previous projects.

haseeb-heaven: I have seen this [GPT4ALL Voice for GPT](https://github.com/Ai-Austin/GPT4ALL-Voice-Assistant) project thought it would be useful here.

mrdiamonddirt: Yeah that project is using pyttsx3 and speech recognition with SpeechRecognition and whisper which I've only now just got running on my system 😅 but yeah probably the way to go ☺️ it's seems much faster at loading the base model also compared to everything else I've tried

haseeb-heaven: Or if you find any other project which is faster and reliable than this you can use it. 
Good luck implementing this into open-interpreter.

mrdiamonddirt: much simpler tts with pytts and speech recognition with whisper as prommised still need to add code to check os or highjack the already existing fuctions

interpreter/voice.py
```diff
@@ -5,6 +5,7 @@ class SpeechAssistant:
     def __init__(self, wakeup_word="interpreter"):
```
haseeb-heaven, line 5: Please add try-catch to this and print error if debug is turned on

mrdiamonddirt:
added dependencies to poetry

KillianLucas:
Incredible implementation, and one of the most futuristic PRs we have (and it was done so early on!)

Closing because I think we should scope down OI as much as possible, and voice seems like an app folks would build on top of the OI core. (If I had it my way, even the terminal interface would be a separate package.)

Issue #326, opened 2023-09-13
Customizable Model Directory


tb3256:
### Is your feature request related to a problem? Please describe.

My main drive on C:\ is running out of space so i need to store the models on a different drive

### Describe the solution you'd like

Allow us to set the directory of our models and where they are downloaded with a new --model-dir [directory] command line tag and arg

Joackk:
Yes, I also need him. It seems that their default value will be installed in the "appdata\local" directory, and I need to be able to specify his path

douglas-hirsh:
If you want to run a model from anywhere you can use the following python file from within a python virtual environment that has the requirements installed.

```python
import interpreter
from llama_cpp import Llama

model_path = "absolute path to your model file"
n_gpu_layers = 40
debug_mode = True
context_window = 16384

llama_2 = Llama(model_path=model_path, n_gpu_layers=n_gpu_layers, verbose=debug_mode, n_ctx=context_window)

interpreter.local = True
interpreter.llama_instance = llama_2
interpreter.chat("Write the code for a fastapi endpoint for todos that supports all CRUD operations. Write all the code in a file called todo_api.py.") # Executes a single command
interpreter.chat() # Starts an interactive chat
```

jordanbtucker:
This feature is being considered. For now, you can create a symlink to the directory of your choosing and Open Interpreter will honor that.

fynnfluegge:
Would also appreciate this. I already have these models on my system so it rather shouldn't be downloaded twice. Symlink is a good workaround for now. I have an idea how to solve this in a generic way with best user experience. Will raise a PR rather sooner than later 🙌

KodeMunkie:
Symlink not really a workaround for Windows machines.

ericrallen:
Closing this as it is a duplicate of #294.

PR #343, opened 2023-09-14
Add custom model directory and download error handling

michaelzdrav:
As mentioned in [Issue 326](https://github.com/KillianLucas/open-interpreter/issues/326), there is a use-case of allowing the user to specify a different path than the default given path. 

The user should be able to select different paths, and as long as it is a directory on the same filesystem or a symbolic link, it will work.

### DIFF
--- a/interpreter/get_hf_llm.py
+++ b/interpreter/get_hf_llm.py
@@ -111,38 +111,50 @@ def get_hf_llm(repo_id, debug_mode, context_window):
     else:
         # If the file was not found, ask for confirmation to download it
         download_path = os.path.join(default_path, selected_model)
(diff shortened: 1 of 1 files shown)


KillianLucas:
Hi @michaelzdrav! This is a great feature idea and I think people really want this, but I think we should provide this as an option in a `config.yaml`, or passed in via `--local_llm_path` or something like that. The reason = we want folks to be able to set this in Python, so IMO it's best to not use the CLI wizard for most things.

(tbh in general I think the CLI wizard is already too powerful, and we should move more things to python-compatible stuff like command line arguments + config.yaml)

PR #356, opened 2023-09-14
Backport support for Python 3.9

jordanesikati:
Work with either Python 3.9 or Python 3.10, or any compatible versions within those ranges.


jordanesikati:
**$ pip install open-interpreter** could not work for me, because **Python 3.9** was installed on my computer, instead of 3.10. 
so I decided to add it as a dependency, solving the issue.

With these modifications, I installed open-interpreter from my fork, and it worked perfectly.
**$pip install git+https://github.com/jordanesikati/open-interpreter.git**

jordanbtucker:
I think we can support Python 3.9. All dependencies either support 3.8 or newer or they just don't specify. We can't support 3.8 because we use `ast.unparse` which requires 3.9, but that seems like the only case.

I just need three changes to merge this:
- Rebase against the current main branch.
- Change the version specifier to `^3.9`. We don't need `^3.9 || ^3.10` since `^3.9` means `>=3.9.0 && <4.0.0`.
- Run `poetry lock --no-update`.

Let me know if you need help with any of that. Otherwise, I can just edit your PR.

@KillianLucas Do you see any issues I'm missing with allowing 3.9? We can also drop support for it in the future if we need to.

jordanesikati:
Hi @jordanbtucker,  all the suggested tasks have been completed.

1- Rebase against the current main branch ... **done**
2- Change the version specifier to ^3.9 ...  **done**
3- Run poetry lock --no-update ... **done**

I haven't faced any issues during the process.

jordanesikati:
Hi @jordanbtucker,
You can handle further adjustments.
Regards,
J. Sikati

ericrallen:
Hey there, @jordanesikati!

Sorry for the long delay on this one.

If you want to resolve the merge conflicts or open a new PR that makes this change, I'd be happy to 👍 it and get it into the merge queue.

ericrallen:
Hey there folks!

I'm going to close this one for now, as testing with Python `3.9` seems to indicate that there is something being used by Open Interpreter or one of its dependencies that is incompatible with `3.9` and we'd need to do more than just change the version in the `pyproject.toml` file to add support.

PR #365, opened 2023-09-14
Pull request template created

TanmayDoesAI:
Created a pull request template


jordanbtucker:
Great idea. Let me see if we can use the YAML form templates for this instead. They are much better than markdown templates in my opinion and they prevent users from just erasing the template.

jordanbtucker:
Okay, doesn't look like we can use YAML templates for this, so the markdown looks good.

PR #369, opened 2023-09-15
Update GPU.md

l-Shane-l:
I needed to add the --user flag to avoid getting permission errors when installing, otherwise the it would complain about not having access and the following step in the docs would return false.


jordanbtucker:
Thanks for the suggestion. While I understand that your situation required use of the `--user` flag, not all user will have that requirement. If `llama-cpp-python` is installed via the `--user` flag but `open-interpreter` is not, this can cause issues as well.

Since not many users have reported this issue, I'm going to close this PR for now. Soon we will likely be recommending `pipx` which will solve this issue completely.

Issue #358, opened 2023-09-14
Model Selection prompt bug  🐞 (Code LLaMa)


haseeb-heaven:
### Describe the bug

Model selection prompt still comes even you select the model via command line arguements.
While selecting model from `--model` flag it still asks us to select the model.

### Reproduce

1. Go to terminal.
2. Type `interpreter --local --model TheBloke/CodeLlama-7B-Instruct-GGUF`
3. See model prompt is still asking to choose paramter for CodeLlama.

### Expected behavior

It should directly set the model without prompting us again and again because we already mentioned in arguments.

### Screenshots

<img width="1116" alt="image" src="https://github.com/KillianLucas/open-interpreter/assets/11544739/4ca4e229-35df-43df-858f-03ae290076db">

### Open Interpreter version

0.1.4

### Python version

3.11.1

### Operating System name and version

Mac OS 13.5 (Ventura)

haseeb-heaven:
> What I think the issue is that you have added "--local" as a flag. I think removing it would solve the problem [Readme](https://github.com/KillianLucas/open-interpreter#running-open-interpreter-locally)

<img width="1338" alt="image" src="https://github.com/KillianLucas/open-interpreter/assets/11544739/d296a27e-2cad-4d8d-92e2-78044d453f28">
It's still giving me prompt of selecting the model size which should be set through flag as well. This bug is only for models which are not GPT.

jordanbtucker:
@haseeb-heaven This is because the `--model` are accepts a Hugging Face model name, which is a repository of files.

For example, if you visit https://huggingface.co/TheBloke/CodeLlama-13B-Instruct-GGUF/tree/main you'll see there are many different gguf files in that repository. Open Interpreter doesn't know which one you want to use, so it prompts you.

haseeb-heaven:
> @haseeb-heaven This is because the `--model` are accepts a Hugging Face model name, which is a repository of files.
> 
> For example, if you visit https://huggingface.co/TheBloke/CodeLlama-13B-Instruct-GGUF/tree/main you'll see there are many different gguf files in that repository. Open Interpreter doesn't know which one you want to use, so it prompts you.

Yes i know that but is there any option to make this interpreter remember which model i have set like in config file because i already downloaded medium size model but it still asking me to choose the model again and again which is annoying?

jordanbtucker:
> is there any option to make this interpreter remember which model i have set

Not at this time, but you're welcome to create a feature request.

PR #396, opened 2023-09-16
Remembering Local Models for Faster Access

haseeb-heaven:
This pull request addresses a bug where the system repeatedly prompts the user to select a local model, even after it has been downloaded once. The issue arises due to the lack of persistent storage for the model information. To enhance user experience and streamline local model usage, this PR introduces a solution that involves saving the Model Name and Model Path to a JSON file. By doing so, we ensure that the system remembers the locally downloaded model, eliminating the need for redundant user input and making local model usage faster and more user-friendly.

This is the new config file that is saved into local machine for local model information.
```json
{
    "config": [
        {
            "model_name": "TheBloke/CodeLlama-7B-Instruct-GGUF",
            "model_path": "/Users/haseeb-mir/codellama-7b-instruct.Q4_K_M.gguf"
        }
    ]
}
```

This is the issue mentioned here.
[Issue#358](https://github.com/KillianLucas/open-interpreter/issues/358)</br>

I have reviewed my code with proper error and exception handling and for test cases for every situation to handle them and it passes them.

- [x] MacOS

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -164,8 +164,12 @@ class Interpreter:
                               if config['model_name'].strip() and config['model_path'].strip():
                                   return config['model_name'], config['model_path']
           return None, None
-      except:
-          return None, None
+      except Exception as e:
+          print(f"An error occurred while loading the model configuration: {str(e)}")
+          if self.debug_mode:
+              traceback.print_exc()
+          error = traceback.format_exc()
+          raise Exception(error)
       
 
   def get_info_for_system_message(self):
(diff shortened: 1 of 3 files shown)


interpreter/interpreter.py
```diff
@@ -130,6 +131,43 @@ def cli(self):
...
+                  configs = json.load(file)['config']
+                  if configs:
+                      for config in configs:
+                          if 'model_name' in config and 'model_path' in config:
+                              # Check if model_name and model_path are not empty strings
+                              if config['model_name'].strip() and config['model_path'].strip():
+                                  return config['model_name'], config['model_path']
+          return None, None
+      except:
+          return None, None
```
moming2k, line 168: would you add the error message if exception happened, otherwise the silent exception would make it difficult to debug by user.

interpreter/interpreter.py
```diff
@@ -315,15 +352,28 @@ def chat(self, message=None, return_messages=False):
     # ^ verify_api_key may set self.local to True, so we run this as an 'if', not 'elif':
     if self.local:
 
+      # Load the model configuration
```
moming2k, line 355: should this code block (line 356-364) moves below line 366, which will only trigger when self.llama_instance == None:

interpreter/interpreter.py
haseeb-heaven: Okay would add error message here

interpreter/interpreter.py
haseeb-heaven: Okay i updated this logic to load if llama model is None.
It tries to load from configuration first if model was not found then it will try to install it.

redthing1:
Please merge this

haseeb-heaven:
> Please merge this

This one needs to be reviewed by someone first

redthing1 (approved):

PR #405, opened 2023-09-16
Update code_interpreter.py

### LINKED ISSUE #123: Feature Request: ChatGPT Plus sub use
This project is wonderful! You're amazing!

I would like to put it on the back burner to consider enabling the use of GPT-4 through a ChatGPT+ subscription via headless chrome scrapping or something like that.

I realize this is a very different direction than API interaction, but it would be so lovely for cost reasons. :)

gagebt:
### Reference any relevant issue (Fixes #123)

This fixes invalid syntax code runs by LLM crashing the interpreter

- [x] Windows
- [ ] MacOS
- [ ] Linux

### DIFF
--- a/interpreter/code_interpreter.py
+++ b/interpreter/code_interpreter.py
@@ -448,45 +448,55 @@ class AddLinePrints(ast.NodeTransformer):
         return new_node
 
 def add_active_line_prints_to_python(code):
-    """
-    Add print statements indicating line numbers to a python string.
-    """
-    tree = ast.parse(code)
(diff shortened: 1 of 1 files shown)


ericrallen:
Hey there, @gagebt!

It looks like this needs to be adjusted to account for recent changes to the codebase.

If you think this is still helpful, could you describe the changes and resolve the merge conflicts. There was a major refactor, so the code needs to be shifted around to account for that.

ericrallen:
I’m going to close this one as stale.

Issue #414, opened 2023-09-17
Custom Instructions


yonaka43:
### Is your feature request related to a problem? Please describe.

Sometimes the code generated by Open Interpreter is deprecated. I understand that Open Procedures is doing a very good job, but what about the idea of allowing users to give custom instructions to Open Interpreter?

### Describe the solution you'd like

Enter **Custom Instruction Mode** with the `--instructions` flag. Users have prepared custom instructions in `.md` format in the `instructions` folder. When generating system messages, it will include relevant instructions from this folder.

### Additional context

If you're interested, I'm ready to submit a pull request. I would be honored to help make this great project even greater.

FrancyJGLisboa:
That's a must-have feature

jfischburg-us:
I've been rather amazed by Gorilla when it comes to determining API's to access: https://gorilla.cs.berkeley.edu/
This might be a really solid integration/plugin to consider incorporating.

MikeBirdTech:
`custom_instructions` can be set in profiles. Thanks!

PR #421, opened 2023-09-18
Feature/instructions

jusanz:

It seems that features like custom instructions are not yet available in Open Interpreter. This feature would allow engineers to prepare detailed, purpose-specific instructions in advance, making it possible for users who are not well-versed in coding to benefit from Open Interpreter.

### Reference any relevant issue (Custom Instructions #414)

[Custom Instructions](https://github.com/KillianLucas/open-interpreter/issues/414#issue-1899720663)

We will place custom instructions in Markdown format (`*.md`) in the `instructions` folder. When launching the interpreter with the `--instructions` flag, the system message will include custom instructions relevant to the content of the query.

I think this would be useful when you want to avoid clearly deprecated code or when you want to execute original commands.

- [x] Windows
- [x] MacOS
- [x] Linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -239,3 +239,6 @@ misc/
 
 # Ignore litellm_uuid.txt
 litellm_uuid.txt
+
+# Ignore instructions
+instructions/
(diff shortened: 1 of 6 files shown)


jusanz:
A major update is reportedly planned! I will close this pull request for now.

PR #431, opened 2023-09-19
Handle %load_message failure in interpreter.py

richawo:
### Issue

When the path for the %load_message doesn't exist, it currently errors out and causes Open Interpreter to come to a hard stop. 

### Proposed Solution

I've prevented the error by simply adding a condition to catch the exception and printing a message explaining the failure.

- [✅ ] I have performed a self-review of my code:

- [ ] Windows
- [✅ ] MacOS
- [ ] Linux

### AI Language Model
- N/A

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -287,10 +287,13 @@ class Interpreter:
       json_path = "messages.json"
     if not json_path.endswith(".json"):
       json_path += ".json"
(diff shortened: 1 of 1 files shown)


mathiasrw:
LGTM

KillianLucas:
Thanks @richawo, great commit! Would be annoying for it to break on this. Perfect fix.

PR #452, opened 2023-09-20
Retry connecting to openai after hitting rate limit to fix #442

### LINKED ISSUE #442: Rate limit will kill the process
### Describe the bug

If we get an `openai.error.RateLimitError: Rate limit reached` error the process will halt. I consider this a bug because it was described as fixed as part of #46 

### Reproduce

- Make a new account with openAI
- Start using the interpreter with lots of recurring challenges (edit x and try to run until there are no errors)

### Expected behavior

The program pauses for a bit of time, informs me that it hit a rate limit, and then it tries again shortly after so the ratelimit does not halt the current process

### Screenshots

image

### Open Interpreter version

0.1.4

### Python version

3.11.5

Mac 12.6

### Additional context

This can be dealt with in interpreter/interpreter.py", line 621 by wrapping in a loop or try catch that deals with rerunning the request after a few seconds if (and only) an exception is based on a rate limit

mathiasrw:
Retry connecting to openai after hitting rate limit to fix #442

Make sure to keep the process running even if we encounter a rate limit in the openAI api. 

Will fix #442 

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [x] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### How it looks

image

--- 

image

Note how this process would have been lost if I had not been running on code that included this PR.

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -24,6 +24,7 @@ from .message_block import MessageBlock
 from .code_block import CodeBlock
 from .code_interpreter import CodeInterpreter
 from .get_hf_llm import get_hf_llm
+from openai.error import RateLimitError 
 
 import os
 import time
@@ -599,11 +600,14 @@ class Interpreter:
 
     # Make LLM call
     if not self.local:
+      
       # GPT
-      
+      max_attempts = 3  
+      attempts = 0  
       error = ""
-      
-      for _ in range(3):  # 3 retries
+
+      while attempts < max_attempts:
+        attempts += 1
         try:
 
             if self.use_azure:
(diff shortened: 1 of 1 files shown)


KillianLucas:
Simple and clean. Thanks @mathiasrw, great contribution!

interpreter/interpreter.py
```diff
@@ -636,7 +640,11 @@ def respond(self):
                 )
 
             break
-        except:
+        except RateLimitError as rate_error:  # Catch the specific RateLimitError
+            print(Markdown(f"> We hit a rate limit. Cooling off for {attempts} seconds..."))
+            time.sleep(attempts)  
+            max_attempts += 1
```
brauliolomeli, line 646: Shouldn't it be `attempts += 1`? Seems like `attempts < max_attempts` will be always True if `attempts` is not incremented.

PR #471, opened 2023-09-21
Update MACOS.md

ihgalis:
We should add this warning to prevent readers from this document to open issues when trying to use the GPU with Intel CPUs in their Macs.


ericrallen (approved):
This doc might be a little outdated in general, but it's good to have something that specifies that we're talking about non-Intel Macs.

Thanks!

PR #476, opened 2023-09-22
MOD: close subprocess and its file handles when reset

coderfengyun:

Problem I encountered: When I use the same Interpreter instance to complete multiple code-writing tasks in a row within a process, I try to use Interpreter.reset() to reset the context and Interpreter state between each task. However, after several runs, I ran into the problem of too many open file handles.

Solution I provide: So in this pr, Interpreter.reset closes the child process and closes its open file handles.

- [x] Windows
- [x] MacOS
- [x] Linux

- [x] GPT4
- [x] GPT3
- [x] Llama 7B
- [x] Llama 13B
- [x] Llama 34B
- [x] Huggingface model (Please specify which one)

### DIFF
--- a/interpreter/code_interpreters/subprocess_code_interpreter.py
+++ b/interpreter/code_interpreters/subprocess_code_interpreter.py
@@ -35,6 +35,8 @@ class SubprocessCodeInterpreter(BaseCodeInterpreter):
 
     def terminate(self):
         self.process.terminate()
+        self.proc.stdin.close()
+        self.proc.stdout.close()
 
     def start_process(self):
         if self.process:


AncalagonX:
Would this also be desired for %undo if there were new handle openings that were undone?

coderfengyun:
> Would this also be desired for %undo if there were new handle openings that were undone?

I think if the code-interpreter subprocess is already started before %undo, is a toss-up, depending on preferences for code. 
1. Pros: clearing has the obvious benefit of releasing the file handle ASAP
2. Cons: not clearing speeds up the processing of the next question in this chat session, and the file handle is eventually released when the chat session is closed or reset.

So, a decision needs to be made by the owner of this code repository.

interpreter/code_interpreter.py
```diff
@@ -117,6 +117,15 @@ def start_process(self):
...
                      daemon=True).start()
+    
+  def stop(self):
+    """
+    Stops the interpreter process.
+    """
+    if self.proc:
+      self.proc.terminate()
+      self.proc.stdin.close()
+      self.proc.stdout.close()
```
ericrallen, line 128: I think these changes probably need to move to [`interpreter/core/core.py`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/core/core.py) now that the code has been refactored.

interpreter/interpreter.py
```diff
@@ -192,6 +192,11 @@ def reset(self):
     Resets the interpreter.
     """
     self.messages = []
+    
+    #遍历并stop每一个code interpreter
+    for code_interpreter in self.code_interpreters.values():
+      code_interpreter.stop()
```
ericrallen, line 198: I think this needs to move to [`interpreter/core/core.py`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/core/core.py) now that the codebase has been refactored.

ericrallen:
Hey there, @coderfengyun!

I think this is a really great suggestion and am happy to help resolve the merge conflicts and move the code to the refactored codebase if you want to hand it off.

I think we just need to add the `stdin` and `stdout` calls to `.close()` to the existing [`terminate()` handler](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/code_interpreters/subprocess_code_interpreter.py#L36).

interpreter/code_interpreter.py
coderfengyun: Of course, I'll come back to modify this commit.

coderfengyun: What do you think about turning off stdin and stdout directly in the terminate method?

coderfengyun:
> Hey there, @coderfengyun!
> 
> I think this is a really great suggestion and am happy to help resolve the merge conflicts and move the code to the refactored codebase if you want to hand it off.
> 
> I think we just need to add the `stdin` and `stdout` calls to `.close()` to the existing [`terminate()` handler](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/code_interpreters/subprocess_code_interpreter.py?rgh-link-date=2023-10-24T20%3A36%3A58Z#L36).

Changes were made as per your REVIEW comments.

KillianLucas:
Great work @coderfengyun, very straightforward and helpful! Merging now, then renaming `.proc` to `.process` which is I believe the right property name. 

Long time coming and I very much appreciate your edits post-refactor. Thank you @ericrallen for your work here too.

coderfengyun:
Thanks to all the contributors to open-interpreter for giving us the opportunity to explore the use of code-interpreter. What I did was make a modest contribution.

PR #24, opened 2023-09-02
feat: add semgrep code scanning and guarddog for package installations

ericrallen:
This PR introduces some tools to help verify that code _might_ be safe to execute.

It adds a `--scan` flag that can be used to enable code scanning via [semgrep](https://semgrep.dev/) and a `--guarddog` flag that can be used to enable scanning [PyPI](https://github.com/DataDog/guarddog#pypi) and [npm](https://github.com/DataDog/guarddog#npm) packages with [guarddog](https://github.com/datadog/guarddog) before installing them.

`--scan` has 3 possible values: `auto` (always scan code with semgrep before asking to execute), `ask` (ask the user if they want to scan a code snippet before asking to execute), and `off` (don't ask to scan code).

Both the `--scan` and `--guarddog` options are disabled by default and are entirely opt-in.

This is a super interesting project, and while I know this won't make it 100% safe to execute Open Interpreter code, I hope it can start to provide some extra assurance and help mitigate some common threats.

Once [datadog/guarddog #249](https://github.com/DataDog/guarddog/issues/249) is resolved we can make some slight adjustments and include `guarddog` in the `pyproject.toml` dependencies, but for now the version of semgrep that it is pinned at has issues with scanning individual files like this with the automatic configuration based on [returntocorp/semgrep-rules #2894](https://github.com/returntocorp/semgrep-rules/issues/2894)

Just as a note, I was actually impressed that while testing some vulnerable files that _should_ trigger semgrep's rules Open Interpreter tried to save me from myself:

>  I'm sorry, but executing this code is not safe. The code contains potentially harmful operations, such as reading the /etc/passwd file, which contains sensitive system information. It's important to only execute code when you're sure it's safe and understand what it does.
> 
> If you need to test this code, I recommend doing so in a controlled environment, such as a virtual machine or a container, where it can't cause harm to your actual system.

I had to let it know that I was working on a code scanning implementation for Open Interpreter and needed it to try to execute this kind of code for testing purposes.

PR #484, opened 2023-09-23
feat: add semgrep code scanning via -safe argument

ericrallen:

This PR introduces the first tool in a safety toolkit to help verify that code _might_ be safe to execute.

It adds a `-safe` argument that can be used to enable code scanning via [semgrep](https://semgrep.dev/).

`-safe` has 3 possible values:

- `off`: (_default_) don't enabled `safe_mode`
- `auto`: enable `safe_mode` and always scan code with `semgrep` before asking to execute
- `ask`: enable `safe_mode` and ask the user if they want to scan a code snippet with `semgrep` before asking to execute

**Note**: The `-safe` option is disabled by default and is entirely opt-in. Enabling `safe_mode` disables `auto_run`.

### Reference any relevant issue (Replaces #24)

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [x] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

I tested this with code generated and read from the filesystem with `gpt-3.5-turbo` and `gpt-4`, but the code scanning doesn't include the model in any way, so it should function regardless of what model the user has configured.

### DIFF
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -73,6 +73,14 @@ arguments = [
         "nickname": "ak",
         "help_text": "optionally set the API key for your llm calls (this will override environment variables)",
         "type": str
+    },
+    {
+        "name": "scan_code",
+        "nickname": "scan",
+        "help_text": "optionally scan code with semgrep",
+        "type": str,
+        "default": "off",
+        "choices": ["off", "ask", "auto"]
     }
 ]
 
@@ -85,7 +93,10 @@ def cli(interpreter):
         if arg["type"] == bool:
             parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], action='store_true', default=None)
         else:
-            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"])
+            choices = arg["choices"] if "choices" in arg else None
+            default = arg["default"] if "default" in arg else None
+
+            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"], choices=choices, default=default)
 
     # Add special arguments
     parser.add_argument('--config', dest='config', action='store_true', help='open config.yaml file in text editor')
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -93,10 +93,7 @@ def cli(interpreter):
         if arg["type"] == bool:
             parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], action='store_true', default=None)
         else:
-            choices = arg["choices"] if "choices" in arg else None
-            default = arg["default"] if "default" in arg else None
-
-            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"], choices=choices, default=default)
+            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"])
 
     # Add special arguments
     parser.add_argument('--config', dest='config', action='store_true', help='open config.yaml file in text editor')
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -75,9 +75,9 @@ arguments = [
         "type": str
     },
     {
-        "name": "scan_code",
-        "nickname": "scan",
-        "help_text": "optionally scan code with semgrep",
+        "name": "safe_mode",
+        "nickname": "safe",
+        "help_text": "optionally enable safety mechanisms like code scanning; valid options are off, ask, and auto",
         "type": str,
         "default": "off",
         "choices": ["off", "ask", "auto"]
@@ -137,9 +137,9 @@ def cli(interpreter):
         if attr_value is not None and hasattr(interpreter, attr_name):
             setattr(interpreter, attr_name, attr_value)
 
-    # if auto_run is enabled, we won't bother scanning code
-    if interpreter.auto_run and not interpreter.scan_code == "off":
-        setattr(interpreter, "scan_code", "off")
+    # if safe_mode and auto_run are enabled, safe_mode disables auto_run
+    if interpreter.auto_run and not interpreter.safe_mode == "off":
+        setattr(interpreter, "auto_run", False)
 
     # Default to CodeLlama if --local is on but --model is unset
     if interpreter.local and args.model is None:
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -140,6 +140,10 @@ def cli(interpreter):
         if attr_value is not None and hasattr(interpreter, attr_name):
             setattr(interpreter, attr_name, attr_value)
 
+    # if auto_run is enabled, we won't bother scanning code
+    if interpreter.auto_run and not interpreter.scan_code == "off":
+        setattr(interpreter, "scan_code", "off")
+
     # Default to CodeLlama if --local is on but --model is unset
     if interpreter.local and args.model is None:
         # This will cause the terminal_interface to walk the user through setting up a local LLM
--- a/interpreter/code_interpreters/language_map.py
+++ b/interpreter/code_interpreters/language_map.py
@@ -1,0 +1,17 @@
+from .languages.python import Python
+from .languages.shell import Shell
+from .languages.javascript import JavaScript
+from .languages.html import HTML
+from .languages.applescript import AppleScript
+from .languages.r import R
+
+
+language_map = {
+    "python": Python,
+    "bash": Shell,
+    "shell": Shell,
+    "javascript": JavaScript,
+    "html": HTML,
+    "applescript": AppleScript,
+    "r": R,
+}
--- a/interpreter/code_interpreters/languages/python.py
+++ b/interpreter/code_interpreters/languages/python.py
@@ -4,6 +4,8 @@ import ast
 import re
 
 class Python(SubprocessCodeInterpreter):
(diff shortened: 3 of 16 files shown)


interpreter/cli/cli.py
```diff
@@ -84,7 +92,10 @@ def cli(interpreter):
         if arg["type"] == bool:
             parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], action='store_true')
         else:
-            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"])
+            choices = arg["choices"] if "choices" in arg else None
+            default = arg["default"] if "default" in arg else None
+
+            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"], choices=choices, default=default)
```
ericrallen, line 98: There wasn't a way to add choices and defaults with the current code, but I didn't want to add more conditional branches, so this is what I came up with. Happy to adjust this to make it more Pythonic if there's a better approach.

interpreter/code_interpreters/language_map.py
```diff
@@ -0,0 +1,17 @@
+from .languages.python import Python
+from .languages.shell import Shell
+from .languages.javascript import JavaScript
+from .languages.html import HTML
+from .languages.applescript import AppleScript
+from .languages.r import R
+
+
+language_map = {
```
ericrallen, line 9: I ended up needing to reference the same language map that the `create_code_interpreter` module was using, so I moved it into it's own module to make it easier to share.

ericrallen:
Screen Shot 2023-09-23 at 12 33 40 AM

interpreter/cli/cli.py
```diff
@@ -124,6 +135,10 @@ def cli(interpreter):
         if attr_value is not None and hasattr(interpreter, attr_name):
             setattr(interpreter, attr_name, attr_value)
 
+    # if auto_run is enabled, we won't bother scanning code
+    if interpreter.auto_run and not interpreter.scan_code == "off":
+        setattr(interpreter, "scan_code", "off")
```
ericrallen, line 140: Not sure if there's a better pattern to handle interconnected attributes like this, but I'm guessing there might be a few things in the future that might depend on the status of `auto_run`, so we might want to explore a less naive approach than what I've done here.

interpreter/core/core.py
```diff
@@ -30,6 +30,7 @@ def __init__(self):
         self.auto_run = False
         self.debug_mode = False
         self.max_output = 2000
+        self.scan_code = "off"
```
ericrallen, line 33: I wasn't sure if we'd want to put this under some sort of larger `security`/`safety` subsection so that the `config.yaml` is a little easier to understand and doesn't become too polluted with top-level keys as more features (like using `guarddog` to scan dependencies before installing them) are added.

Maybe something like:

```yaml
system_message: |
  You are Open Interpreter, a world-class programmer that can complete any goal by executing code.
  ...
local: false
model: "gpt-4"
temperature: 0
security:
  scan_code: "ask"
```

But I didn't want to start fiddling with the config parsing logic if I didn't need to.

interpreter/code_interpreters/languages/python.py
```diff
@@ -4,6 +4,8 @@
 import re
 
 class Python(SubprocessCodeInterpreter):
+    file_extension = "py"
```
ericrallen, line 7: These `file_extension`s aren't strictly necessary, but it felt nicer when generating the temporary file for semgrep to scan to do it with the appropriate extension and could come in handy for some future ideas - like adding some functionality to directly export generated code into a file via a some `%magic` command without needing to have another exchange with Open Interpreter to get it to create some code to pipe the other code into the file.

In the case of something like the JavaScript language configuration this could get a little more complicated since they could also be modules that use the `mjs` extension, but for the purposes it's being used for right now - a naive extension suffices.

ericrallen:
**NOTE**: This depends on #511 to work as expected.

KillianLucas:
@ericrallen this is brilliantly, beautifully done. To me this is a strong foundation for a `--safe` flag. 

I'm in favor of fewer, simpler options so I'm going to remove the tiers. `--safe` will simply scan the code automatically (let me know if this is very important for certain workflows, or if `semgrep` takes a very long time which would justify the tiers). We can add additional security features into it later.

Just to confirm before merging this (and #511), users don't have to log in to semgrep to use it?

ericrallen:
@KillianLucas I've cleaned this one up a bit given the desire to put safety tools under a single flag in the future - I've also simplified the output and made `safe_mode` take precedence over `auto_run` given that automatically running code is the opposite of safe.

I moved the temporary file creation and cleanup into it's own utility since there are other safety tools, like scanning for exposed secrets, that will likely need to leverage a similar temporary file functionality.

Semgrep's [free tier](https://semgrep.dev/pricing/) works really well for this usecase - just scanning a snippet of code locally - but the CLI also allows users who have a Semgrep account to login and leverage some of the more advanced features if they want.

Here's some recordings of the output for each value of the flag:

### Open Interpreter with `-safe "off"`:
OpenInterpreter-SafeMode-Off

### Open Interpreter with `-safe "ask"`:
OpenInterpreter-SafeMode-Ask

### Open Interpreter with `-safe "auto"`:
OpenInterpreter-SafeMode-Auto

ericrallen:
Here are the prompts I've been using to quickly test this functionality:

## Generate Safe (hopefully) Code

```
Solve FizzBuzz for the numbers 0 through 17 with a basic Python loop and if statement. No need to explain the rules of the game, just show us the code.
```

## Generate Vulnerable Code

```
We just added a code scanner to Open Interpreter that relies on Semgrep to detect vulnerable code. Please generate a Python script with an obvious vulnerability, like SQL injection for example, that Semgrep should be able to identify. DO NOT MENTION THE VULNERABILITY IN THE SCRIPT VARIABLE NAMES OR CODE COMMENTS. This script should look like a developer accidentally included the vulnerability without realizing it. DO NOT WARN US ABOUT THE VULNERABILITY OR ITS IMPLICATIONS. We want to see if the code scanner identifies the vulnerability without you alerting us to it.
```

ericrallen:
You can also save the following snippets to files on your machine and ask Open Interpreter to read and execute the code.

**Note**: These examples were generated by Open Interpreter as examples of vulnerable code that should be identified by the code scanner.

## JavaScript

```js
const vm = require('vm');
const contextObject = { globalVar: 1 };

// safe
vm.runInContext('globalVar *= 2;', contextObject);

// vulnerable
let userInput = 'this.constructor.constructor("return process.env")()'; // Value supplied by user input
vm.runInContext(`globalVar = ${userInput};`, contextObject);

// safe
const code = `return 'hello ' + name`
vm.compileFunction(code, [], { parsingContext: vm.createContext({ name: 'name' }) })

// vulnerable
let userInput = '1; while (true)
```

## Python

```py
import subprocess
import sys

# Vulnerable
user_input = "foo && cat /etc/passwd" # value supplied by user
subprocess.call("grep -R {} .".format(user_input), shell=True)

# Vulnerable
user_input = "cat /etc/passwd" # value supplied by user
subprocess.run(["bash", "-c", user_input], shell=True)

# Not vulnerable
user_input = "cat /etc/passwd" # value supplied by user
subprocess.Popen(['ls', '-l', user_input])

# Not vulnerable
subprocess.check_output('ls -l dir/')
```

ericrallen:
@KillianLucas I rebased this branch, fixed a bit of code that I forgot to update from `scan_code` to `safe_mode`, and added a loading indicator during the scan so that the user has some feedback that things are happening. This used to display the `semgrep` output, but that had a bit too much noise and it felt like it needed something to let the user know that things are happening.

## Loading Indicator

OpenInterpreter-SafeMode-Auto-Loading

Also here's a simple prompt for anyone who wants to test this to get right into a state where you can test the scanner after running with `poetry run interpreter -safe "ask"` or `poetry run interpreter -safe "auto"`:

> Solve FizzBuzz for 0 through 17. Don't explain the code or tell me your process or how FizzBuzz works. Just generate the code so we can execute it.

interpreter/code_interpreters/languages/python.py
```diff
@@ -4,6 +4,9 @@
...
+    proper_name = "Python"
```
ericrallen, line 8: These proper names are mostly just being used for the code scanner to say

> Code Scanner: No Issues were found with this Python code

We could just rely on the lowercase language string from the code block, but it felt nicer and more human-readable this way.

interpreter/terminal_interface/terminal_interface.py
```diff
@@ -8,13 +8,25 @@
...
-        display_markdown_message("""**Open Interpreter** will require approval before running code. Use `interpreter -y` to bypass this.
+        interpreter_intro_message = [
+            "**Open Interpreter** will require approval before running code."
+        ]
+
+        if interpreter.safe_mode != "off":
+            interpreter_intro_message.append(f"**Safe Mode**: {interpreter.safe_mode}")
+        else:
+            interpreter_intro_message.append(
+                "Use `interpreter -y` or set `auto_run: true` to bypass this."
```
ericrallen, line 24: Since `safe_mode` is opt-in and disables `auto_run` it seemed unnecessary to mention the `-y` flag.

ericrallen:
Not sure why the workflow tests are failing. Everything seems to work when I run the test suite locally.

ericrallen:
@KillianLucas I’ll rebase again tomorrow to resolve the `poetry.lock`conflict, and try to see if I can get to the bottom of what’s wrong with the test suite, but this PR only applies features behind a flag, so I’m not sure how it impacted them.

Any guidance on getting it ready to merge would be appreciated.

KillianLucas:
GREAT WORK ERIC! Literally our most impressive and extensive PR on this whole project to date. Extreme utility here, and you've opened the door to a substantial wing of the OI project (and maybe even to an industry/discipline??) focused on making LLM-written code **safe**. Merging now. 🎉

PR #508, opened 2023-09-25
fix: stop overwriting boolean config values

ericrallen:
### Describe the changes you have made:

Without the default set to `None`, any boolean CLI flag that isn't passed manually via `-` or `--` reverts to its`False` even if it is configured in the `config.yaml` file.

### Reference any relevant issue (Fixes #507 )

- [x] I have performed a self-review of my code:

### I have tested the code on the following OS:
- [ ] Windows
- [x] MacOS
- [ ] Linux

### AI Language Model (if applicable)
- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B
- [ ] Huggingface model (Please specify which one)

PR #509, opened 2023-09-25
feat: add guarddog option for scanning packages

ericrallen:

This adds a `--guarddog` flag that instructs the model to use [guarddog](https://github.com/datadog/guarddog) to scan any `npm` or `pip` packages before installing them.

### Reference any relevant issue (Replaces #24)

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

**NOTE**: This depends on #508 in order to allow the `guard_dog` setting in the `config.yaml` to work as expected.

### DIFF
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -72,7 +72,13 @@ arguments = [
         "nickname": "ak",
         "help_text": "optionally set the API key for your llm calls (this will override environment variables)",
         "type": str
-    }
+    },
+    {
+        "name": "guarddog",
+        "nickname": "guarddog",
+        "help_text": "scan PyPI and npm packages for malicious code with guarddog",
+        "type": bool,
+    },
 ]
 
 def cli(interpreter):
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -30,6 +30,7 @@ class Interpreter:
         self.auto_run = False
         self.debug_mode = False
         self.max_output = 2000
+        self.guarddog = False
 
         # Conversation history
         self.conversation_history = True
--- a/interpreter/core/respond.py
+++ b/interpreter/core/respond.py
@@ -6,6 +6,7 @@ from ..rag.get_relevant_procedures import get_relevant_procedures
 from ..utils.truncate_output import truncate_output
 import traceback
 import litellm
+import platform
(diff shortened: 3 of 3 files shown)


KillianLucas:
Hi @ericrallen, thank you so much for this contribution and for your initial contribution earlier this month. I think this is an important step towards securing Open Interpreter.

I think the docker requirement is too heavy. But I think we should map out the best way to do this comprehensively once we get past that. `--guarddog` to me is not an intuitive flag, but I've thought a lot about `--safe_mode` or something like that. This should do several things to protect the user IMO (virtual environment, for example). 

For these reasons I'll close this PR but would really like to merge once we've sketched out an intuitive, comprehensive `--safe`-ish flag.

I can definitely see the use of guard-dog at that point once the docker requirement is fully lifted cross-platform. Then we can programmatically scan the imported packages, more like your first commit, instead of relying on the LLM.

Thank you for thinking about this. It's critical, maybe the most critical thing we can be working on.

CodeAKrome:
Is it possible to use either this or a filter like it for things like grabbing web pages? Some kind of middleware perhaps.

ericrallen:
@CodeAKrome `guarddog` is purely focused on scanning a PyPI or npm packages for threats, but I do have another feature branch in the works that will integrate another tool that scans text for prompt injection and a few other things.

ericrallen:
@KillianLucas I don't think there will be an option to run `guarddog` on Windows without Docker, but we could try to tell Open Interpreter to look for the Windows Subsystem for Linux and run `guarddog` there and/or to check to see if guarddog is available and offer to scan packages before installing them.

I'll take a look at adding a slightly modified approach to the `safe_mode` toolkit after we get #484 merged in.

I think scanning packages prior to installing them will be super important, especially for users who aren't familiar with the package or the dependency ecosystem for the given language.

We can also look into something like [OWASP's `dependency-check`](https://owasp.org/www-project-dependency-check/) and [GitLab's Package Hunter](https://about.gitlab.com/blog/2021/07/23/announcing-package-hunter/), but I think Package Hunter requires a slightly more complex setup process.

PR #511, opened 2023-09-25
fix: allow args to have choices and defaults

### LINKED ISSUE #510: Non-boolean Args Can't Define Choices or Defaults
### Describe the bug

Arguments that aren't boolean can't currently define choices or defaults because they aren't passed to `parser.add_argument()`

### Reproduce

1. Add a new argument to the `arguments` array in `./interpeter/cli/cli.py`
2. Give your argument a `choices` array and a `default` value
3. Note that the default value is ignored

### Expected behavior

Choices and defaults are properly handled.

### Screenshots

_No response_

### Open Interpreter version

0.1.4

### Python version

3.11.2

MacOS 10.15.7

### Additional context

Discovered while working on #484 

Will remove from that PR so that it can be addressed separately as other work depends on it, too.

ericrallen:

This allows non-boolean args to define possible options and default values, which were ignored previously.

### Reference any relevant issue (Fixes #510)

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

Companion PR to #508 discovered while working on adding a new feature.

### DIFF
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -84,7 +84,10 @@ def cli(interpreter):
         if arg["type"] == bool:
             parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], action='store_true')
         else:
-            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"])
+            choices = arg["choices"] if "choices" in arg else None
+            default = arg["default"] if "default" in arg else None
+
+            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"], choices=choices, default=default)
 
     # Add special arguments
     parser.add_argument('--config', dest='config', action='store_true', help='open config.yaml file in text editor')


interpreter/cli/cli.py
```diff
@@ -84,7 +84,10 @@ def cli(interpreter):
         if arg["type"] == bool:
             parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], action='store_true')
         else:
-            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"])
+            choices = arg["choices"] if "choices" in arg else None
+            default = arg["default"] if "default" in arg else None
+
+            parser.add_argument(f'-{arg["nickname"]}', f'--{arg["name"]}', dest=arg["name"], help=arg["help_text"], type=arg["type"], choices=choices, default=default)
```
ericrallen, line 90: There wasn't a way to add `choices` and `defaults` with the current code, but I didn't want to add more conditional branches, so this is what I came up with.

Happy to adjust this to make it more Pythonic if there's a better approach.

KillianLucas:
Merging so we can merge #484.

PR #521, opened 2023-09-25
fix %load_message

CodeAKrome:
interpreter.load() missing.
Assuming it was just missed

- [ ] Windows
- [ ] MacOS
- [x] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -114,3 +114,5 @@ class Interpreter:
         for code_interpreter in self._code_interpreters.values():
             code_interpreter.terminate()
         self._code_interpreters = {}
+    def load(self, messages):
+        self.messages = messages


KillianLucas:
Hi @CodeAKrome, I figured we should take this out since `interpreter.messages = messages` feels more intuitive to me, same length code, and will sort of show people how the deeper `interpreter` object works (remembers messages in `interpreter.messages`) which might help people understand how to extend it. I def should have documented this decision, so thank you for contributing to restore this feature.

What do you think? Should we add it back? If any workflows are significantly improved by it, I could understand that.

KillianLucas:
Closing for the reasons mentioned above, but I'm very open to restoring `interpreter.load()` if there's a good reason. Thanks!

PR #522, opened 2023-09-26
Fix GNU readline support for user input

cowile:

This commit adds an import for the readline module at the top of terminal_interface.py to provide GNU readline support for user input. This will allow Unix users to edit their input and browse their command history like in a UNIX shell. The readline module is imported conditionally, so the script does not break for Windows users where this module is typically not available.

pyproject.toml already mentions the ability to do this when declaring the pyreadline dependency and interpreter.py also imports readline, but it doesn't work unless the import is in the same module using the input function.

And I must credit my co-developer Open Interpreter for identifying the correct file and making the change at my direction.

- [X] I have performed a self-review of my code:

- [ ] Windows
- [ ] MacOS
- [X] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/terminal_interface/terminal_interface.py
+++ b/interpreter/terminal_interface/terminal_interface.py
@@ -3,6 +3,11 @@ The terminal interface is just a view. Just handles the very top layer.
 If you were to build a frontend this would be a way to do it
 """
 
+try:
+    import readline
+except ImportError:
+    pass
+
 from .components.code_block import CodeBlock
 from .components.message_block import MessageBlock
 from .magic_commands import handle_magic_command


KillianLucas:
Hi @cowile, thank you for noticing this! I think we should drop the dependency since even the existence of the package caused problem for Windows users. Until we can more thoroughly test it across platforms, I'm not sure if we should try to add it back.

What do you think? Open Interpreter had this until a few days ago -- let me know if it was beneficial to your workflow and then yes, I think a conditional import would be great.

cowile:
@KillianLucas If the pyreadline dependency is causing problems for Windows users then I agree to dropping it.

I can't say pyreadline specifically has benefited my workflow as I have not attempted to use open-interpreter on Windows yet. But I will strongly argue that a line editing feature is a minimum expectation for usability. No one wants to backspace 50 times if they made a typo at the beginning of the line.

The conditional import I added should definitely go in as readline is a system library on Unix-like platforms, no 3rd party dependency needed.

KillianLucas:
Oh shit how do we do line editing? Is that a readline thing? That would be awesome. 

Ideally we'd have full blown editor stuff like the ability to paste in multiline strings, not sure if that's possible. I'll look into this but if you know already, let me know!

cowile:
"Line editing" is a generic term for the functionality provided by libraries like readline. Mostly it means moving around and editing the line you've typed before submitting it.

readline is powerful and customizable. By default, pressing the enter key triggers the readline action accept-line, but that is configurable. You can make any other key accept-line and have the enter key type an ordinary newline character that can be edited along with the rest. See [here](https://stackoverflow.com/questions/161495/is-there-a-nice-way-of-handling-multi-line-input-with-gnu-readline). readline supports [configuration files](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File.html). Looks like python [gives](https://docs.python.org/3/library/readline.html) you `readline.read_init_file` to read whatever file you choose change much of readline's behavior without any code changes.

ericrallen:
I'd live to give a huge `+1` to this PR.

Not being able to easily edit the prompt and move my cursor is a constant source of frustration and this PR is a marked improvement in quality of life.

KillianLucas:
Yeah, you're both right. I only saw the error that made me remove it on one Windows machine anyway (and not on another + any Mac or Linux system).

`readline` is back!

PR #523, opened 2023-09-26
Update WINDOWS.md

rsfutch77:
Warns the user to re-launch cmd windows after installing llama locally.


KillianLucas:
Makes sense, thanks @rsfutch77!

PR #530, opened 2023-09-26
add custom_llm_provider option for litellm 

kylehh:
1, Add custom_llm_provider which will be used for litellm for custom openai usage (like [Anyscale Endpoint](https://app.endpoints.anyscale.com/landing))
2, Fixed a bug in setup_text_llm which didn't remove the first message. 

- [x ] I have performed a self-review of my code:

- [ ] Windows
- [x ] MacOS
- [ x] Linux

- [ ] GPT4
- [ x] GPT3
- [ x] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -72,6 +72,12 @@ arguments = [
         "nickname": "ak",
         "help_text": "optionally set the API key for your llm calls (this will override environment variables)",
         "type": str
+    },
+    {
+        "name": "custom_llm_provider",
+        "nickname": "cp",
+        "help_text": "custom llm provider name, used for litellm",
+        "type": str
     }
 ]
 
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -46,6 +46,7 @@ class Interpreter:
         self.api_key = None
         self.max_budget = None
         self._llm = None
+        self.custom_llm_provider = None
 
         # Load config defaults
         config = get_config()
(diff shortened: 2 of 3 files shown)


KillianLucas:
I think this should work by setting `--api_base`? Or is this to change the format of the API call away from "openai" (I'm an OpenAI guy so sorry for my confusion!)

kylehh:
Hi @KillianLucas yes, in order to work with other OpenAI API compatible endpoint services, we need to supply both api_base  and custome_llm_provider parameter, and this is due to the implementation of [LiteLLM](https://github.com/BerriAI/litellm/blob/ce62dc40b9f0d2662fb345a690ce14baaa4a45dd/litellm/main.py#L352). 

If you directly call openai.ChatCompletion.create API, then api_base is enough.

KillianLucas:
Hey @kylehh we need better docs on this but it looks like we can skip this parameter [by letting folks pass it in before a "/" in the model name.](https://github.com/BerriAI/litellm/blob/ce62dc40b9f0d2662fb345a690ce14baaa4a45dd/litellm/main.py#L247)

So users should be able to put "custom_openai/" before the model name then set `api_base` to use it as an OpenAI compatible endpoint. Let me know if that works and I'll close this PR + update the docs, if not tho I'll merge it.

(tbh ideally we'd be able to put everything in the model string... i guess we can't because api_base needs to be set separately, but ideally we'd just have one magic string to identify each LLM. for now i think the best we can do is just having as few params as possible)

kylehh:
Hi @KillianLucas Thanks for looking closely into this issue!  I tried your suggestion but unfortunately it's not working for now b/c anyscale is not in llm provider_list yet. I will make a PR to litellm to add it so it should work.
If so, I will remove the adding custom_llm_provider part of the PR and only leave the bug fix in setup_text_llm.py

kylehh:
The changes should be made in litellm. And I will close this PR and open another for the bug fix
Thanks!

PR #545, opened 2023-09-28
Update get_config.py

zmottie:
Updating get_config.py to use the same code to retrieve config file as cli.py. Unintended results (C:\Users\MyUser\AppData\Local\Open Interpreter\Open Interpreter\config.yaml) were happening.

Simplified get_config.py to use the same code as cli.py which returns the correct results.

- [x] Windows

All


KillianLucas:
Great catch @zmottie, thank you so much for finding/fixing this. Merging now, will push to `pip` soon.

PR #554, opened 2023-09-28
fix: stop overwriting safe_mode config.yaml setting with default in args

ericrallen:

`safe_mode` from `config.yaml` was being overwritten by the `default` option set in the arguments config. I forgot that we already have a default value in the `core.py` `Interpreter` instance and the way args were being handled meant that the default value functionality I added in #511 was actually overriding the value from `config.yaml`.

### Reference any relevant issue (Updates #484)

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


KillianLucas:
Nice, great catch Eric! Merging.

PR #560, opened 2023-09-29
Fixed a bug in setup_text_llm.py

kylehh:
Remove the system message from the message list
Otherwise the system message will show up twice

This is similar to the treatment in this openai [case](https://github.com/KillianLucas/open-interpreter/blob/7a3f54d688e46dabde58f698574f300791b1944a/interpreter/llm/setup_openai_coding_llm.py#L48C30-L48C30)

- [ ] Windows
- [ x] MacOS
- [ x] Linux

- [ ] GPT4
- [ ] GPT3
- [ x] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


KillianLucas:
Oh, duh. Thank you so much for catching this @kylehh! Merging now.

PR #566, opened 2023-09-30
Magic command defaults upgrade + delete messages

### LINKED ISSUE #564: Messages shouldn't overwrite by default
### Is your feature request related to a problem? Please describe.

```
def handle_save_message(self, json_path):
    if json_path == "":
      json_path = "messages.json"
    if not json_path.endswith(".json"):
      json_path += ".json"
    with open(json_path, 'w') as f:
      json.dump(self.messages, f, indent=2)

    display_markdown_message(f"> messages json export to {os.path.abspath(json_path)}")
```

Right now, the messages overwrite by default when no path is provided. That shouldn't be the case. You could lose conversations by mistake - especially without warnings in place.

### Describe the solution you'd like

I'd prefer that the saved by date, and an incremental suffix.

### Describe alternatives you've considered

Considered just added a warning to prevent overwriting but that is a lot of friction for every time you want to save quickly. The best user experience is storing a new json, especially since they aren't large files it's fine. We can even add a simple cleanup command if needed.

The warning can still be added but this will only be when you type in the same file name.

### Additional context

_No response_

richawo:
### The problem:
Magic commands previously saved to and overwrote message.json by default. The problem is:

1. You want to save messages quickly, but you don't necessarily want it to overwrite.
2. The name message.json is uninformative - you know nothing about the contents.
3. The message JSONs are also small, so storing multiple past conversations easily outweighs the cons.

### Relevant issue
Fixes #564 

### The Fix:

1. Updated the save name convention to be `[date]_messages_[counter].json` when no path is provided.
2. Loads the latest message `[date]_messages_[counter].json` when no path is provided.
3. Added %delete_messages to let you clean up historical messages, which accepts a parameter (for the number of files you wish to preserve) - or deletes everything except the most recent file by default.
4. Updated the `handle_help` function to reflect the changes.

With this approach, you can still save the messages quickly without overwriting. The names also include the date so you're able to get more information about the file (albeit nothing too in-depth). If you wish to clear out previous messages, you can also do that now too.

### Testing
- [ ✅] I have performed a self-review of my code:

- [ ] Windows
- [✅ ] MacOS
- [ ] Linux

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,8 @@ We welcome code contributions through pull requests. Here are some guidelines:
 
 - Open a PR to `main` linking any related issues. Provide detailed context on your changes.
 
+- Add your `OPENAI_API_KEY` environment variable to the repository secrets to make sure your tests can pass.
+
 - We will review PRs when possible and work with you to integrate your contribution. Please be patient as reviews take time. 
 
 - Once approved, your code will be merged - thank you for improving Open Interpreter!
--- a/interpreter/terminal_interface/magic_commands.py
+++ b/interpreter/terminal_interface/magic_commands.py
@@ -1,6 +1,7 @@
 from ..utils.display_markdown_message import display_markdown_message
 import json
 import os
+import datetime
 
 def handle_undo(self, arguments):
     # Removes all messages after the most recent user entry (and the entry itself).
@@ -82,7 +83,13 @@ def default_handle(self, arguments):
 
 def handle_save_message(self, json_path):
     if json_path == "":
-      json_path = "messages.json"
+      now = datetime.datetime.now()
+      msg_counter = 0
+      while True:
+        json_path = now.strftime("%Y-%m-%d") + "_messages_" + str(msg_counter) + ".json"
+        if not os.path.exists(json_path):
+          break
+        msg_counter += 1
     if not json_path.endswith(".json"):
       json_path += ".json"
     with open(json_path, 'w') as f:
--- a/interpreter/terminal_interface/magic_commands.py
+++ b/interpreter/terminal_interface/magic_commands.py
@@ -2,6 +2,7 @@ from ..utils.display_markdown_message import display_markdown_message
 import json
 import os
 import datetime
+import glob
 
 def handle_undo(self, arguments):
     # Removes all messages after the most recent user entry (and the entry itself).
@@ -98,8 +99,13 @@ def handle_save_message(self, json_path):
     display_markdown_message(f"> messages json export to {os.path.abspath(json_path)}")
 
 def handle_load_message(self, json_path):
-    if json_path == "":
-      json_path = "messages.json"
+    if not json_path:
+      json_paths = glob.glob("*_messages_*.json")
+      if json_paths:
+        latest_path = max(json_paths, key=os.path.getmtime)
+        json_path = latest_path
+      else:
+        json_path = "messages.json"
     if not json_path.endswith(".json"):
       json_path += ".json"
     with open(json_path, 'r') as f:
--- a/interpreter/terminal_interface/magic_commands.py
+++ b/interpreter/terminal_interface/magic_commands.py
@@ -40,8 +40,9 @@ def handle_help(self, arguments):
       "%debug [true/false]": "Toggle debug mode. Without arguments or with 'true', it enters debug mode. With 'false', it exits debug mode.",
       "%reset": "Resets the current session.",
       "%undo": "Remove previous messages and its response from the message history.",
-      "%save_message [path]": "Saves messages to a specified JSON path. If no path is provided, it defaults to 'messages.json'.",
-      "%load_message [path]": "Loads messages from a specified JSON path. If no path is provided, it defaults to 'messages.json'.",
+      "%save_message [path]": "Saves messages to a specified JSON path. If no path is provided, it defaults to '[date]_messages_[counter].json'.",
+      "%load_message [path]": "Loads messages from a specified JSON path. If no path is provided, it loads the latest default message '[date]_messages_[counter].json'.",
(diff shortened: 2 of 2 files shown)


richawo:
Took a look, no reason for builds to fail. None of my code touches the interpreter logic - purely magic commands.

I've also tested it, and all works fine. Having looked through some other PRs (open and closed) seems to be a recurring issue.

richawo:
image

Evidence of testing

CONTRIBUTING.md
```diff
@@ -31,6 +31,8 @@ We welcome code contributions through pull requests. Here are some guidelines:
 
 - Open a PR to `main` linking any related issues. Provide detailed context on your changes.
 
+- Add your `OPENAI_API_KEY` environment variable to the repository secrets to make sure your tests can pass.
+
```
ericrallen, line 35: I think this should be removed as it's a bit confusing, and will likely lead to folks opening Issues or reaching out in the Discrd because they are trying to add their own OpenAI keys to a repo that they don't have access to the Environment Variables for - and I don't think anyone is running the actions on their own forks of the repo.

interpreter/terminal_interface/magic_commands.py
```diff
@@ -38,8 +40,9 @@ def handle_help(self, arguments):
       "%debug [true/false]": "Toggle debug mode. Without arguments or with 'true', it enters debug mode. With 'false', it exits debug mode.",
       "%reset": "Resets the current session.",
       "%undo": "Remove previous messages and its response from the message history.",
-      "%save_message [path]": "Saves messages to a specified JSON path. If no path is provided, it defaults to 'messages.json'.",
-      "%load_message [path]": "Loads messages from a specified JSON path. If no path is provided, it defaults to 'messages.json'.",
+      "%save_message [path]": "Saves messages to a specified JSON path. If no path is provided, it defaults to '[date]_messages_[counter].json'.",
+      "%load_message [path]": "Loads messages from a specified JSON path. If no path is provided, it loads the latest default message '[date]_messages_[counter].json'.",
```
ericrallen, line 44: This is a minor rendering thing, but I think the underscores (`_`) here need to be escaped with `\_` or they try to render as markdown and end up italicized.

### Before escaping underscores in filenames

Screen Shot 2023-10-08 at 11 55 58 AM

### With filename underscores escaped in `%load_message` help text

Screen Shot 2023-10-08 at 11 57 15 AM

interpreter/terminal_interface/magic_commands.py
```diff
@@ -38,8 +40,9 @@ def handle_help(self, arguments):
...
+      "%delete_message [number]": "Deletes saved messages, preserving the most recent 'number' files. If no number is provided, it deletes all but the most recent file.",
```
ericrallen, line 45: `%delete_message` didn't seem to work. Maybe because it's defined as `delete_messages` on [line 148 in the `handle_magic_command()` method](https://github.com/KillianLucas/open-interpreter/pull/566/files#diff-39f7ea35aad6a3d890783709177df802705530ae6f13d67a43a94656d7b15076R148)?

Screen Shot 2023-10-08 at 12 02 04 PM

Also, it seems like `%delete_message` should require a path rather than potentially deleting multiple files when called without an argument.

ericrallen:
Overall, I really like this change. It's a solid improvement on storing and retrieving conversation threads. Nice work!

I wonder if we should use the same path that the user's `config.yaml` gets stored in if they don't provide a path - rather than polluting whatever current directory they were in when they initiated `interpreter`

I believe getting the correct directory for the user's `config.yaml` file is handled in [`interpreter/utils/get_config.py`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/utils/get_config.py#L10) and we could use similar logic or extract that path creation into a separate utility that this functionality can import, too.

interpreter/terminal_interface/magic_commands.py
richawo: Good spot - will update!

CONTRIBUTING.md
richawo: That's fair, I thought it was the reason for the failing tests initially but I realised it wasn't. Will delete.

Would be good to get to the bottom of it more generally since it's a recurring issue..

interpreter/terminal_interface/magic_commands.py
richawo: Yeah, it should be %delete_messages, and will make it so that you can supply a folder or file.

richawo:
> I wonder if we should use the same path that the user's `config.yaml` gets stored in if they don't provide a path - rather than polluting whatever current directory they were in when they initiated `interpreter`
> 
> I believe getting the correct directory for the user's `config.yaml` file is handled in [`interpreter/utils/get_config.py`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/utils/get_config.py#L10) and we could use similar logic or extract that path creation into a separate utility that this functionality can import, too.

Yeah, I'm happy to extract this out if we're re-using the logic across multiple places.

ericrallen:
@richawo I think just making the update so that `%delete_messages` works and resolving the most recent merge conflicts would be enough, and we can look at abstracting the common functionality out into a utility in a separate PR.

Thanks for taking this one on!

richawo:
@ericrallen, hey, I'll sort it out shortly! Sorry, I've been distracted by too many projects lately 😅

ericrallen:
No worries, @richawo.

I just wanted to revisit this one because I think it's a nice quality-of-life improvement.

Happy to help with the resolution of Merge Conflicts, too. I know sometimes they can get a bit tricky after so many changes have been made since a PR started.

MikeBirdTech:
Hi @richawo 
Any desire to revive this PR? Or should I close it?
Thanks!

richawo:
Haha I really should, still not fixed right?

MikeBirdTech:
@richawo 
haha even though a lot has changed, it still defaults to `messages.json` if no path is provided.

Let me know if you have any questions or any other ideas for how to improve magic commands!

PR #571, opened 2023-10-02
Fix bug when trying to use local non-CodeLlama model

alexweberk:
Python split method and indexing into the second element in the list does not work when the string pattern is not found; there would be only one element in the resulting list.

changed from using `split` method to using `replace` method.

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B
- [x] Huggingface model (Please specify which one): TheBloke/Mistral-7B-Instruct-v0.1-GGUF


KillianLucas:
Oh my god of course. Dealt with this exact thing earlier. Thanks for catching @alexweberk!

PR #582, opened 2023-10-04
Ability to add external functions

guiramos:
I am making use of OI as a fundamental component in my project and these changes are extremely welcome and increases OI flexibility in my opinion. Please consider this carefully. 

This PR will provide 2 changes:

1 - mechanism to add more functions to the API call, By default, the code function will be inserted as it is currently defined. 
The process involves 2 steps. Adding the additional function schemas to the list that will be passed to the API and their handling functions. As per API definition, only one function can be executed at a time.
The original schema definition of "execute" function was not changed. The execution of the function was moved the `code_function_handler.py` file 

2 - Making the `relevant procedures` optional. I found this to be not very useful for my application, maybe other will think similarly. So, my proposal is to make it optional. Again, by default, the behavior is the same as it was.

- [ ] Windows
- [X] MacOS
- [ ] Linux

- [X] GPT4
- [X] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


guiramos:
@KillianLucas 

I've been trying to keep this PR as much updated as possible. I last MERGE from main resulted in the conflict in both respond.py and setup_openai_coding_llm as they are HEAVILY hard-coded for the single function "code".

I believe in an extensible implementation and I've been testing this one successfully with 3 external functions working with the native "code" one.

Please reach out to me if you want to discuss the behavior or another design.

guiramos:
Merged with main branch now. It's getting harder and harder to maintain this.

Please help incorporate such functionality.

MikeBirdTech:
Hi @guiramos 
I apologize that this PR didn't get reviewed in a timely manner. The codebase has changed drastically in that time and I suggest we close this PR as it won't align with the current architecture without a massive re-write. Thank you for putting your time and effort into this

PR #594, opened 2023-10-05
chore: update test suite

ericrallen:

This adds a `system_message` prepending test, a `messages.reset()` test, and makes the math testing a little more robust while also trying to prevent some edge cases where the llm would respond with explanations or an affirmative "Sure I can do that. Here's the result..." or similar responses instead of just the expected results of a test.

It also simplifies some of the testing setup/teardown by introducing a `setup_function` and `teardown_function` instead of repeating things like `interpreter.reset()` in every test.

I noticed while testing locally that sometimes the assistant wouldn't respond, but I definitely wasn't hitting any rate limits. Introducing a slight delay via  `time.sleep()` into the process helped.

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/tests/test_interpreter.py
+++ b/tests/test_interpreter.py
@@ -1,26 +1,94 @@
+from random import randint
 import interpreter
+import time
+
 interpreter.auto_run = True
 interpreter.model = "gpt-3.5-turbo"
 interpreter.temperature = 0
 
-def test_hello_world():
+
+# this function will run before each test
+# we're clearing out the messages Array so we can start fresh and reduce token usage
+def setup_function():
     interpreter.reset()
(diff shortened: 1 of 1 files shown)


KillianLucas:
Oh, absolutely. No brainer merge— we need this bad. Great work Eric.

---

Related: In the future we should explore testing OI's ability to create files, use basic APIs— more OI-specific abilities. **Eventually, with different system messages and language models.** Changes to hard-coded messages + LLM integration methods = literal shots in the dark without this.

I guess ^this is more benchmark stuff. World's first code interpreter obstacle course / benchmark.

(I'm planning on getting deep into that after the hackathon next week, but if you wanted to give it a shot I'd merge 👍)

PR #595, opened 2023-10-06
Add num_gpu_layers and num_threads for better inference of local models

hxjerry:
Weaker hardware may lack vram and require part of the inference to be done on cpu. Passing num_threads and num_gpu_layers would be greatly beneficial.

- [ ] Windows
- [ ] MacOS
- [x] Linux

- [ ] GPT4
- [ ] GPT3
- [x] Llama 7B
- [x] Llama 13B
- [x] Llama 34B


interpreter/cli/cli.py
```diff
@@ -80,6 +80,18 @@
         "help_text": "optionally enable safety mechanisms like code scanning; valid options are off, ask, and auto",
         "type": str,
         "choices": ["off", "ask", "auto"]
+    },
+    {
+        "name": "threads",
```
ericrallen, line 85: This argument never actually gets passed to the `interpreter` instance because `interpreter.threads` doesn't exist.

In [`core.py`](https://github.com/KillianLucas/open-interpreter/pull/595/files#diff-44ef24ada573d406ea9f66d0669046a162d58d6db6bbd83308f4704b7eedb5f7R35) the `Interpreter` Class attribute is called `num_threads`, so the argument parsing logic drops the `--threads` passed to the CLI because it doesn't find a corresponding attribute.

ericrallen:
Hey there!

I think if the actual properties we're trying to send to the model are `n_threads` and `n_gpu_layers`, we should probably call the properties we're adding `n_threads` and `n_gpu_layers` to be explicit and avoid any confusion about what these values are and where they are used.

hxjerry:
Fixed bug & names, also resolved conflict, should be ready to merge now.

ericrallen:
Hey there!

Sorry for the delay, I’ve been traveling for work and forgot to circle back to this one.

If you want to reopen this one with the conflicts resolved or start a fresh branch without any conflicts, ping me and I’ll be happy to approve your PR.

PR #597, opened 2023-10-06
[Example] Basic plugin system functionality via interpreter.plugins entrypoint

ericrallen:

This allows plugins to be created by developers and distributed via PyPI, installed by users via `pip`, and then loaded by Open Interpreter dynamically if the user adds the plugin name to their `config.yaml` file.

Currently this allows plugins to extend the functions array given to the LLM.

An example web scraper plugin is available at: https://github.com/InterwebAlchemy/open-interpreter-plugin-demo

The magic that allows plugins to register with Open Interpreter is handled in their own [`pyproject.toml`](https://github.com/InterwebAlchemy/open-interpreter-plugin-demo/blob/main/pyproject.toml#L12) via `tool.poetry.plugin."interpreter.plugins"`.

**Note**: Currently Open Interpreter executes the function, but I'm not sure how to get it responding based on data in the function's results.

#### Web Scraping Plugin Example

OpenInterpreter-Plugins-with-Data

### How to test

1. `gh pr checkout https://github.com/KillianLucas/open-interpreter/pull/597`
2. `poetry install`
3. `poetry shell`
4. `pip install https://github.com/InterwebAlchemy/open-interpreter-plugin-demo`
**Note**: You may need to clone this one and `poetry build` it locally first and then `pip install /path/to/cloned/repo`; I'm not a Python dependency expert.
5. Add `browse` plugin to your `config.yaml` in the `plugins` Array:
    ```yaml
    local: false
    model: "gpt-4"
    temperature: 0
    system_message: # your system message here
    debug_mode: false
    safe_mode: "ask"

    plugins:
      - browse
    ```
6. `poetry run interpreter` 
7. Ask it something like this:
> What are the top news stories on https://lite.cnn.com/ right now?

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/llm/setup_openai_coding_llm.py
+++ b/interpreter/llm/setup_openai_coding_llm.py
@@ -3,36 +3,17 @@ from ..utils.merge_deltas import merge_deltas
 from ..utils.parse_partial_json import parse_partial_json
 from ..utils.convert_to_openai_messages import convert_to_openai_messages
 from ..utils.display_markdown_message import display_markdown_message
+from ..plugins.function_manager import extend_functions
 import tokentrim as tt
 
-
-function_schema = {
-  "name": "execute",
-  "description":
-  "Executes code on the user's machine, **in the users local environment**, and returns the output",
-  "parameters": {
-    "type": "object",
-    "properties": {
-      "language": {
-        "type": "string",
-        "description":
-        "The programming language (required parameter to the `execute` function)",
-        "enum": ["python", "R", "shell", "applescript", "javascript", "html"]
-      },
-      "code": {
-        "type": "string",
-        "description": "The code to execute (required)"
-      }
-    },
-    "required": ["language", "code"]
-  },
-}
 
 def setup_openai_coding_llm(interpreter):
     """
     Takes an Interpreter (which includes a ton of LLM settings),
     returns a OI Coding LLM (a generator that takes OI messages and streams deltas with `message`, `language`, and `code`).
     """
+
+    functions = extend_functions(interpreter)
 
     def coding_llm(messages):
         
@@ -68,7 +49,7 @@ def setup_openai_coding_llm(interpreter):
             'model': interpreter.model,
             'messages': messages,
             'stream': True,
-            'functions': [function_schema]
+            'functions': functions
         }
 
         # Optional inputs
--- a/interpreter/llm/setup_openai_coding_llm.py
+++ b/interpreter/llm/setup_openai_coding_llm.py
@@ -21,7 +21,7 @@ def setup_openai_coding_llm(interpreter):
         messages = convert_to_openai_messages(messages)
 
         # Add OpenAI's recommended function message
-        messages[0]["content"] += "\n\nOnly use the function you have been provided with."
+        messages[0]["content"] += "\n\nOnly use the functions you have been provided with."
 
         # Seperate out the system_message from messages
         # (We expect the first message to always be a system_message)
@@ -77,6 +77,7 @@ def setup_openai_coding_llm(interpreter):
         accumulated_deltas = {}
         language = None
         code = ""
+        accumulated_arguments = {}
 
         for chunk in response:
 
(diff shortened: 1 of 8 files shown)


ericrallen:
I think an approach like this is preferable to something like #582 because it's easier for developers to extend and easier for users to decide which extended functions they want to make available to Open Interpreter.

interpreter/plugins/base_plugin.py
ericrallen, line 1: This is sort of what the base plugin would look like.

I think we'd want it published in a separate package like `open-interpreter-plugin` so that it's easy for developers to import it into their plugins and then make sure their plugin exports an instance of the `InterpreterPlugin` class with registered functions.

ericrallen:
I've updated the demo plugin to actually scrape a provided URL and updated Open Interpreter to actually execute the supplied plugin's functions, but I'm not quite sure how to make interpreter understand and respond with insights from the results of the function.

interpreter/llm/setup_openai_coding_llm.py
```diff
@@ -118,21 +100,35 @@ def coding_llm(messages):
...
+        if "function_call" in accumulated_deltas and accumulated_deltas["function_call"]["name"] != "execute" and accumulated_deltas["function_call"]["name"] in interpreter.functions:
+            # Execute the function
+            function_name = accumulated_deltas["function_call"]["name"]
+            parameters = accumulated_arguments
+            function = interpreter.functions[function_name]
+            results = function(function_name=function_name, parameters=parameters)
+
+            # Yield the result
+            if results:
+                yield { "message": results }
```
ericrallen, line 132: I'm guessing this probably isn't the best way to handle the function call results, but I'm not quite sure how they should be integrated back into Open Interpreter's messages and given an appropriate response based on the initial request that triggered the function call.

guiramos: Hey @ericrallen, with this implementation, the tests are passing? 
There are two very simple tests there that will break if these changes aren't made correctly.

Please take a look at my changes:
https://github.com/KillianLucas/open-interpreter/pull/582/files#diff-ea1396dac3dbdd50b5e7e15bbee9773b5da17175622ca1bbbf47490670a0a66d

It took me some time to make it dynamic and generic enough to working with basically any function and any amount of additional functions, while keep the original behavior intact.

It seems that @KillianLucas hasn't had enough time to evaluate none of these PRs lately, and offering two implementations to perform conceptually the same thing is not helpful. We should really unify this approach instead of providing challenging implementations.

My initial objective is to provide an extensible way for functions for those who are using OI via code, like mentioned in the description of my PR. So, my "external" (to OI) will always be internal to my project. 

Don't you think that forcing someone/company to publish their plugin to pypip would be an impediment? Not everyone wants to open their code, publishing to pypip publicly. 
I am working on something that is not totally mine, they don't want to open everything.

ericrallen: Hey there!

Thanks for the feedback.

The tests all pass on this branch; I’m guessing that’s because I’m not changing anything about how the existing call to `execute` works.

The logic I’ve implemented doesn’t currently feed the results of the function call back into the conversation so that the model can use them, though. So, this approach definitely needs some more work and should probably try to borrow more from your example.

This PR is a Draft because I wanted to show a proof of concept and an exploration of how we _could_ provide this functionality rather than a final proposal for how we _should_ provide it.

In terms of how we approach allowing developer’s to extend the functionality, I think leveraging entry points provides a better developer experience for developers who want to create and distribute this kind of functionality and a better end user experience for users who want to consume this additional functionality.

In terms of public vs private extensions, the great thing about relying on a package manager like `pip` for this is that you can install these plug-in packages from a local directory, private GitHub repo, private PyPI registry, public GitHub repo, or the public PyPI registry. As long as the package is available in your environment and makes use of the `interpreter.plugins` entry point, it will be available.

So, you don’t have to release any of your plug-in code if you don’t want to.

Ultimately, I think neither my approach nor yours are likely to be the final implementation, and I actually hope that we can find a way to introduce these functions and extensions at the Python layer instead of the Large Language Model (LLM) layer because:

1. not all LLMs are tuned for function calling and giving them more options can lead to more confusion for the model
2. we won’t have to spend as much of our limited context window explaining and describing functions
3. having the Python layer manage function execution gives us more control over what happens, how it happens, and when it happens

ericrallen:
Going to close this one for now as it has served its purpose as an example of how we could approach extending Open Interpreter’s functionality.

PR #598, opened 2023-10-06
feat: Enable API Logging with Helicone

Kabilan108:
API requests will be logged via Helicone when the `HELICONE_API_KEY` variable is set

- Added if block to set `litellm.api_base` and `litellm.headers` when HELICONE_API_KEY is provided.

- [x] Windows
- [ ] MacOS
- [x] Linux

- [x] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


interpreter/core/respond.py
```diff
@@ -6,6 +6,14 @@
...
 import traceback
 import litellm
+import os
+
+
+# Enable logging with Helicone
+if os.getenv("HELICONE_API_KEY"):
+    litellm.api_base = "https://oai.hconeai.com/v1"
+    litellm.headers = {"Helicone-Auth": f"Bearer {os.getenv('HELICONE_API_KEY')}"}
+
```
ericrallen, line 16: I like the idea of making it simple to use, but this approach actually breaks the existing `--api_base` parameter functionality and prevents users from being able to also test out a local model or another API proxy.

OpenInterpreter-Helicone

I think we should consider making thrid party integrations like this opt-in via a CLI parameter or a flag in the user's `config.yaml` file that can be overridden.

Maybe adding some sort of `integrations` array to the `config.yaml` and then if the user has `helicone` as an entry in that array, we can enable helicone and try to pick up the API key? It might need to follow something similar to the functionality we use in [`interpreter/terminal_interface/validate_llm_settings.py`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/terminal_interface/validate_llm_settings.py#L53) to make sure we can find the `HELICONE_API_KEY` if Helicone is enabled.

Overall, this a solid idea to add and nice to be able to track Open Interpreter's token usage and other datapoints in the Helicone dashboard.

interpreter/core/respond.py
```diff
@@ -14,21 +22,22 @@ def respond(interpreter):
...
         # Open Procedures is an open-source database of tiny, up-to-date coding tutorials.
         # We can query it semantically and append relevant tutorials/procedures to our system message
         get_relevant_procedures(interpreter.messages[-2:])
         if not interpreter.local:
             try:
-                system_message += "\n\n" + get_relevant_procedures(interpreter.messages[-2:])
+                system_message += "\n\n" + get_relevant_procedures(
+                    interpreter.messages[-2:]
+                )
```
ericrallen, line 36: It looks like your editor might have gotten carried away with some automated formatting.

You might want to go back through and remove these extra changes and then either save without formatting or adjust your settings to only apply formatting to modified code on save instead of the entire file.

ishaan-jaff:
@Kabilan108 we expose callbacks through litellm - you can set a custom callback to log data to Helicone if you'd like: https://docs.litellm.ai/docs/observability/custom_callback

interpreter/core/respond.py
Kabilan108: Thanks for the feedback! I implemented this approach and pushed a new commit!
Right now, I added the validation for integrations in the `setup_llm.py` file, but do you think it would be better to have a separate module for it? I've only added helicone for now, but I am thinking of adding support for other LLM monitoring tools like [LangSmith](https://llm.report/openai) and [llm.report](https://llm.report).

interpreter/core/respond.py
Kabilan108: It seems like that section was removed in subsequent commits.

interpreter/config.yaml
```diff
@@ -14,4 +14,5 @@ system_message: |
   You are capable of **any** task.
 local: false
 model: "gpt-4"
-temperature: 0
\ No newline at end of file
+temperature: 0
+integrations: [helicone]
```
ericrallen, line 18: This might be good to put in an example in the docs, but we won’t want to enable this by default in the core `config.yaml`.

interpreter/core/respond.py
ericrallen, line 1: It’s hard to tell, but I don’t think there are any actual changes to this file that are relevant to this PR.

KillianLucas:
Hey Kabilin! We're now approaching integrations with simple guides like [this one](https://docs.openinterpreter.com/integrations/e2b). Let me know if you'd like to reopen this PR, or make a new one into our docs showing how to integrate Open Interpreter with Helicone. Thanks!

Issue #596, opened 2023-10-06
High LLM Costs


ubuntuyeah:
### Describe the bug

The script should show pricing or warn the user about cost being incurred. OpenAI's billing limits don't work so it is easy to exceed your budget.

### Reproduce

1. Run with GPT4
2. Get REKT

### Expected behavior

Warning about the number of tokens consumed, or ask for a budget

### Open Interpreter version

0.1.7

### Python version

3.11.5

### Operating System name and version

Ubuntu 22.04

### Additional context

Really, I'm broke now, and it's your fault!

Ghostyboy0719:
WTH $50???? I have embedded so many documents and generated so much content but still never crossed more than $30 YTD! What did you run?

jacksongoode:
There are quite a few precautions listed in the readme to letting it run in autopilot.

dagelf:
It definitely needs to display a warning or show the cost incurred with every turn. I vote to keep it open as a feature request. 

I also ran up $20 just in the first 15 minutes, and this after never even reaching $5 a month on `pipegpt` and others even after daily use for a few months now. I can only presume it sends a lot of context with every try.

And I ran up this bill by just trying to get it to run code - not even on autopilot - code that never executed due to it not being able to interpret the end quotes for some reason. 

(The end quotes seem to end up inside the code block, causing errors. Also short outputs never terminate. Not sure how this project got so popular - I presume it works out of the box on OSX but definitely not the case on Linux.)

Because of this I'm taking my Saturday morning to dig into the code - it's nice and portable - I'm just not sure why curses is needed, or if it can be bypassed - I'm sure when I get it, it will make debugging easier, but for now its just a hindrance!

dagelf:
Also concur about openAI billing limits not being honored. I set topups of $5 and a cap at $10 and noticed charges going off my card every few minutes. I shudder to think what would happen to someone who set it up thinking they're safe.

Arrendy:
There is a `max_budget` flag that can be set

`interpreter --max_budget 0.01 # < This is in USD!`

ericrallen:
If you pull down the latest `main` branch, you can test out the `%tokens` magic command that attempts to estimate your token usage and cost based on the current conversation. You can run it at any point and it will give you an estimate of how many tokens will be sent along with the next message you send.

It’s not perfect, but it is an attempt to give more information about token usage and help users understand how many tokens are being used behind the scenes to power the conversation.

ericrallen:
[...]

Issue #584, opened 2023-10-04
I started with interpreter --fast but it cost me a lot and it's very annoying


onigetoc:
### Describe the bug

I started with interpreter --fast but it cost me a lot and it's really annoying ;(
please everytime we start a new session, write down everything

Model set to GPT-4

Open Interpreter will require approval before running code.

Use interpreter -y to bypass this.
**ADD SOMETHING LIKE THIS.**
To use GPT-3.5 Turbo: interpreter -- fast
To use locally: --local

Press CTRL-C to exit.

I'm very angry to pay that much for nothing. If i didn't go to my billing account to check i would finish the day with a $100 bill.
I'm very upset right now.

### Reproduce

interpreter --fast

### Expected behavior

- You should confirm what we use GPT-3.5 Turbo or Locally

### Open Interpreter version

latest

### Python version

latest

### Operating System name and version

Windows 10

### Additional context

Again. I'm very angry to pay that much for nothing. If i didn't go to my billing account to check i would finish the day with a $100 bill.
I'm very upset right now.

All these infos on how to use interpreter should be in the principal Readme page on this repo.
The first time days ago on using it, cost me a lot.

ericrallen:
Hey there!

Sorry you had a frustrating experience.

I can’t help with what’s already happened in the past, but hopefully this can help you and other folks avoid a similar issue in the future:

You can set the default model to `gpt-3.5-turbo` directly in the `config.yaml` file so that you don’t have to think about it again.

I leave mine set that way so that I have to explicitly tell it to use GPT-4 if I think I’ll need it for a specific problem.

Hope that helps for the future!

Just as a note, the README explains [how to choose a model](https://github.com/KillianLucas/open-interpreter#change-your-language-model), but given the rapid pace of change at the moment - and the relatively recent refactor - there’s a chance something is out of sync between messaging from the application itself, the README, and the external documentation site.

onigetoc:
hi, but in all information i had now it say. for gpt-3.5 turbo use : interpreter --fast
and it should work. But if this change without further notice, it's not really helpful.
I think my recommendation was minimal, or using a command to get help and basic information.

ericrallen:
Just to clarify, `interpreter --fast` does use the `gpt-3.5-turbo` model.

![OpenInterpreter-Fast-vs-Regular](https://github.com/KillianLucas/open-interpreter/assets/1667415/5a630849-e851-4064-ae98-01c1e8e2bf18)

[...]

PR #607, opened 2023-10-08
feat: add %tokens magic command that counts tokens via tiktoken

ericrallen:

This adds a `%tokens` magic command that keeps a running tally of how many tokens the current conversation's messages have used and leverages LiteLLM's [`cost_per_token()`](https://docs.litellm.ai/docs/completion/token_usage#2-cost_per_token) method to estimate the cost of the current conversation.

The cost estimate logic can easily be removed if just the token count is preferred

### Without Cost

OpenInterpreter-Tokens-Command

### With Cost Estimate

OpenInterpreter-Tokens-Command-with-Cost

This adds something to help with issues like #596 and #584

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [x] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/utils/count_tokens.py
+++ b/interpreter/utils/count_tokens.py
@@ -1,0 +1,26 @@
+import tiktoken
+
+def count_tokens(text="", model="gpt-4"):
+    """
+    Count the number of tokens in a list of tokens
+    """
+
+    encoder = tiktoken.encoding_for_model(model)
+
+    return len(encoder.encode(text))
+
+
+def count_messages_tokens(messages=[], model=None):
+    """
+    Count the number of tokens in a list of messages
+    """
+
+    tokens_used = 0
+
+    for message in messages:
+        if isinstance(message, str):
+            tokens_used += count_tokens(message, model=model)
+        elif "message" in message:
+            tokens_used += count_tokens(message["message"], model=model)
+
+    return tokens_used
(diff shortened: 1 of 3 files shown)


interpreter/utils/count_tokens.py
```diff
@@ -0,0 +1,38 @@
...
+    Count the number of tokens in a list of messages
+    """
+
+    tokens_used = 0
+
+    for message in messages:
+        if isinstance(message, str):
+            tokens_used += count_tokens(message, model=model)
+        elif "message" in message:
+            tokens_used += count_tokens(message["message"], model=model)
```
ericrallen, line 33: Upon further investigation it seems like this approach might miss the generated code from Open Interpreter when calculating token usage.

I need to tweak things to take that into account as well.

ericrallen: I think this is more accurate now with the recent push.

ericrallen:
I updated this to truncate to six decimal places:

Screen Shot 2023-10-08 at 7 50 45 PM

KillianLucas:
Great PR Eric. Better API cost management and awareness = huge weight lifted for people using this regularly. This will also be great for optimizing the system message.

KillianLucas:
Wait did I merge the right thing in lol? Just did this for the first time from my phone.

PR #609, opened 2023-10-09
feat: add support for loading different config.yaml files

ericrallen:

This adds a `--config_file` option that allows users to specify a path to a config file or the name of a config file in their Open Interpreter config directory and use that config file when invoking `interpreter`.

It also adds similar functionality to the `--config` parameter, allowing users to open and edit different config files by passing the path with `--config_file`.

To simplify finding and loading files, I also added a utility to return the path to a directory in the Open Interpreter config directory and moved some other points in the code from using a manually constructed path to utilizing the same utility method for consistency and simplicity.

I also added a test that uses the `interpreter.extend_config()` method to validate that configuration files can be loaded.

**Note**: This functionality has become more important as I try to help folks with Issues here on GitHub and in the Discord. Because my own `config.yaml` is customized, it's hard to test with the default config without manually renaming/deleting my customized config first.

### Demo

OpenInterpreter-Config-File

### Testing

1. `gh pr checkout https://github.com/KillianLucas/open-interpreter/pull/609`
2. `poetry run interpreter --config --config_file config.testing.yaml` to create a new config file
3. Edit some settings in the config file and save them - `model` is the easiest one to see the effects of
4. `poetry run interpreter --config_file config.testing.yaml` to load your customized config file
5. `poetry run interpreter --config` to see the default config file
6. `poetry run interpreter` to load the default config file
7. `poetry run interpreter --config_file ./tests/config.test.yaml` to load a config file from a relative path (this is the config file that our `pytest` test for config loading uses)

You can also test with local and absolute config file paths.

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -4,7 +4,8 @@ It's the main file. `import interpreter` will import an instance of this class.
 """
 from interpreter.utils import display_markdown_message
 from ..cli.cli import cli
-from ..utils.get_config import get_config
+from ..utils.get_config import get_config, user_config_path
+from ..utils.local_storage_path import get_storage_path
 from .respond import respond
 from ..llm.setup_llm import setup_llm
 from ..terminal_interface.terminal_interface import terminal_interface
@@ -25,6 +26,8 @@ class Interpreter:
         self.messages = []
         self._code_interpreters = {}
 
+        self.config_file = user_config_path
+
         # Settings
         self.local = False
         self.auto_run = False
@@ -35,7 +38,7 @@ class Interpreter:
         # Conversation history
         self.conversation_history = True
         self.conversation_filename = None
-        self.conversation_history_path = os.path.join(appdirs.user_data_dir("Open Interpreter"), "conversations")
+        self.conversation_history_path = get_storage_path("conversations")
 
         # LLM settings
         self.model = ""
@@ -50,14 +53,20 @@ class Interpreter:
         self.gguf_quality = None
 
         # Load config defaults
-        config = get_config()
-        self.__dict__.update(config)
+        self.extend_config(self.config_file)
 
         # Check for update
         if not self.local:
             # This should actually be pushed into the utility
             if check_for_update():
                 display_markdown_message("> **A new version of Open Interpreter is available.**\n>Please run: `pip install --upgrade open-interpreter`\n\n---")
+
+    def extend_config(self, config_path):
+        if self.debug_mode:
+            print(f'Extending configuration from `{config_path}`')
+
+        config = get_config(config_path)
+        self.__dict__.update(config)
 
     def chat(self, message=None, display=True, stream=False):
         if stream:
--- a/interpreter/utils/get_config.py
+++ b/interpreter/utils/get_config.py
@@ -1,25 +1,50 @@
 import os
 import yaml
-import appdirs
 from importlib import resources
 import shutil
 
+from .local_storage_path import get_storage_path
+
 config_filename = "config.yaml"
 
-# Using appdirs to determine user-specific config path
-config_dir = appdirs.user_config_dir("Open Interpreter")
-user_config_path = os.path.join(config_dir, config_filename)
+user_config_path = os.path.join(get_storage_path(), config_filename)
 
-def get_config():
-    if not os.path.exists(user_config_path):
-        # If user's config doesn't exist, copy the default config from the package
-        here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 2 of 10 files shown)


interpreter/core/core.py
```diff
@@ -49,15 +52,21 @@ def __init__(self):
...
             # This should actually be pushed into the utility
             if check_for_update():
                 display_markdown_message("> **A new version of Open Interpreter is available.**\n>Please run: `pip install --upgrade open-interpreter`\n\n---")
 
+    def extend_config(self, config_path):
+        if self.debug_mode:
+            print(f'Extending configuration from `{config_path}`')
+
+        config = get_config(config_path)
+        self.__dict__.update(config)
```
ericrallen, line 68: Moving this into a class method allows it to be called programmatically so that developers can load configs when using `interpreter` in Python scripts, as well.

There's an example of this in [`test_interpreter.py`](https://github.com/KillianLucas/open-interpreter/pull/609/files#diff-d54bbc4862f37675b5d0a8512cf7e7ad8c6d6fbc4f8dcdc3ac5b0e205e90d0f1R31)

tests/test_interpreter.py
```diff
@@ -1,16 +1,18 @@
...
-
 
 # this function will run before each test
 # we're clearing out the messages Array so we can start fresh and reduce token usage
 def setup_function():
     interpreter.reset()
+    interpreter.temperature = 0
+    interpreter.auto_run = True
+    interpreter.model = "gpt-3.5-turbo"
+    interpreter.debug_mode = False
```
ericrallen, line 15: Now that we're testing configuration overrides by loading a custom config file, we need to make sure that our setup function resets to the expected parameters for our tests.

tests/test_interpreter.py
```diff
@@ -54,7 +71,7 @@ def test_hello_world():
         {"role": "assistant", "message": hello_world_response},
     ]
 
-"""
+@pytest.mark.skip(reason="Math is hard")
```
ericrallen, line 74: @KillianLucas I updated this to use `pytest`'s built-in `skip` functionality instead of just commenting it out.

interpreter/core/core.py
```diff
@@ -50,15 +53,21 @@ def __init__(self):
...
-        self.__dict__.update(config)
+        self.extend_config(self.config_file)
 
         # Check for update
         if not self.local:
             # This should actually be pushed into the utility
             if check_for_update():
                 display_markdown_message("> **A new version of Open Interpreter is available.**\n>Please run: `pip install --upgrade open-interpreter`\n\n---")
 
+    def extend_config(self, config_path):
```
ericrallen, line 64: We could also call this `load_config` instead of `extend_config` and then it would be more clear what's going on when you call `interpreter.load_config()` via Python.

interpreter/utils/get_config.py
```diff
@@ -1,25 +1,50 @@
...
+                    path = os.path.join(get_storage_path(), path)
+
+
+                # If user's config doesn't exist, copy the default config from the package
+                here = os.path.abspath(os.path.dirname(__file__))
+                parent_dir = os.path.dirname(here)
+                default_config_path = os.path.join(parent_dir, 'config.yaml')
+
+                # Copying the file using shutil.copy
+                new_file = shutil.copy(default_config_path, path)
```
ericrallen, line 42: I was thinking that we might want to exclude the `system_message` from the new `config.yaml` file in order to have people using the default message as things are upgraded unless they _really_ want to customize it.

Right now any changes we make to the default `system_message` in [`interpeter/config.yaml`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/config.yaml#L1) will be overwritten by the `system_message` in the user's `config.yaml` file, which means if we need to push out some update to the underlying prompt engineering that powers Open Interpreter, users who have already generated a `config.yaml` file will not receive it unless they manually update their `config.yaml` with the latest system prompt.

It might also be good to introduce a property like `append_to_sysem_message` or `appended_system` message that will allow the user to just specify some instructions they want to concatenate onto the end of our default `system_message`.

TanmayDoesAI:
Tested on windows looks good to me, @KillianLucas let's have a check from your end too, this seems too good to not merge!!

interpreter/utils/get_config.py
TanmayDoesAI: Makes sense actually, didn't think of when we would update the system_prompt from our ends.

PR #612, opened 2023-10-10
[WIP] feat: allow editing of code blocks before execution

### LINKED ISSUE #537: Edit code before running
### Is your feature request related to a problem? Please describe.

sometimes the code that is going to run is incorrect, rather than wasting more tokens, I am wasting a lot of tokens for a misplaced brackets.

this issue keeps happening:
```py
f.write('Bootstrap also has advanced ...\n\n')}
```

### Describe the solution you'd like

Any code that is going to run should be saved in a cache file, and the options should be y/n/e
when you type  `e`dit, it should open the file on any default editor and watch for changes, then ask again.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

ericrallen:

This work-in-progress introduces the ability to edit code in your default editor before running it.

When Open Interpreter asks if you want to run the provided code, there is a new option, `%edit`, which will allow you to edit the code.

It currently waits for you to come back and hit `ENTER` to continue execution.

Unfortunately, it seems to automatically run the edited code, and I haven't been able to get it back into the `scan`/`run` loop.

I tried using a few different file/directory watching Python packages, but it seems the file close event is not very easy to listen for and not consistent across platforms, so for the initial proof of concept, I’m relying on user intervention.

#### Demo

OpenInterpreter-Edit-Code

#### Testing Instructions

1. `gh pr checkout https://github.com/KillianLucas/open-interpreter/pull/612`
2. `poetry run interpreter`
   **Note**: `auto_run` must be disabled so don't run it with `-y` or with `auto_run: true` in your `config.yaml`.
3. Provide an input that generates code
   **Example**:
   > Solve FizzBuzz for 0 through 17. Don't explain the code or tell me your process or how FizzBuzz works. Just generate the code so we can execute it.
4. When asked if you want to run the code, enter the `%edit` magic command
5. Edit the code and save your changes
6. Come back to the Terminal and hit `ENTER`

### Reference any relevant issue (Fixes #537)

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -1,10 +1,8 @@
 import argparse
-import subprocess
 import os
-import platform
 import pkg_resources
 import appdirs
-from ..utils.display_markdown_message import display_markdown_message
+from ..utils.open_file import open_file
 from ..terminal_interface.conversation_navigator import conversation_navigator
 
 arguments = [
@@ -114,16 +112,9 @@ def cli(interpreter):
         config_dir = appdirs.user_config_dir("Open Interpreter")
         config_path = os.path.join(config_dir, 'config.yaml')
         print(f"Opening `{config_path}`...")
-        # Use the default system editor to open the file
-        if platform.system() == 'Windows':
-            os.startfile(config_path)  # This will open the file with the default application, e.g., Notepad
-        else:
-            try:
-                # Try using xdg-open on non-Windows platforms
-                subprocess.call(['xdg-open', config_path])
-            except FileNotFoundError:
-                # Fallback to using 'open' on macOS if 'xdg-open' is not available
-                subprocess.call(['open', config_path])
+        
+        open_file(file_path=config_path)
+
         return
     
     # TODO Implement model explorer
--- a/interpreter/terminal_interface/terminal_interface.py
+++ b/interpreter/terminal_interface/terminal_interface.py
@@ -9,6 +9,7 @@ from .magic_commands import handle_magic_command
 from ..utils.display_markdown_message import display_markdown_message
 from ..utils.truncate_output import truncate_output
 from ..utils.scan_code import scan_code
+from ..utils.edit_code import edit_code
 
 
 def terminal_interface(interpreter, message):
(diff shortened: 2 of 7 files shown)


interpreter/terminal_interface/terminal_interface.py
```diff
@@ -119,10 +120,37 @@ def terminal_interface(interpreter, message):
...
+                            ran_code_block = True
+                            render_cursor = False
+
+                            active_block.end()
+
+                            # Add the edited code to the messages
+                            interpreter.messages.append({
+                                **old_message,
+                                "code": edited_code,
+                            })
```
ericrallen, line 151: This doesn't seem to trigger getting back into the same `Do you want to scan this code?` / `Do you want to run this code?` loop and skips straight to execution.

It also doesn't render the code with syntax highlighting, which is a bummer.

interpreter/utils/open_file.py
ericrallen, line 1: This is the same logic that powered `interpreter --config`, but since we're opening other files now, it seemed logical to extract it into a utility that we can call whenever we need to open a specific path in the user's default application.

interpreter/utils/scan_code.py
```diff
@@ -1,34 +1,9 @@
 import os
 import subprocess
 from yaspin import yaspin
-from yaspin.spinners import Spinners
```
ericrallen, line 4: This was just an extraneous import.

interpreter/utils/edit_code.py
```diff
@@ -0,0 +1,64 @@
...
+        print(f"Editing {language_name} code in {file_name}")
+        print("---")
+
+    # Run semgrep
+    try:
+        print("  Press `ENTER` after you've saved your edits.")
+
+        open_file(temp_file)
+
+        with yaspin(text=f"  Editing {language_name} code...").green.right.dots as loading:
```
ericrallen, line 34: There's probably a better way to handle this. The big problem right now is that since we don't really get any kind of useful signal when the user is done, we can't automatically clean up the spinner like we do with code scanning.

This probably needs to shift to an imperative call to a new `yaspin` instance instead of using `with yaspin()` so we have more control over the loading indicator.

interpreter/cli/cli.py
```diff
@@ -1,10 +1,8 @@
 import argparse
-import subprocess
 import os
-import platform
 import pkg_resources
 import appdirs
-from ..utils.display_markdown_message import display_markdown_message
```
ericrallen, line 7: Cleaning up imports that are no longer needed due to the `open_file` utility, and removing `display_markdown_message` as it is no longer used in this file.

ericrallen:
Would love any ideas, suggestions, or assistance getting the edited code block to render as a syntax highlighted code block and ask the user for confirmation before executing.

interpreter/utils/edit_code.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if interpreter.debug_mode:
+        print(f"Editing {language_name} code in {file_name}")
+        print("---")
+
+    # Run semgrep
+    try:
+        print("  Press `ENTER` after you've saved your edits.")
+
+        open_file(temp_file)
```
ericrallen, line 32: Ideally, eventually we'll be able to incorporate some sort of in-place editing that feels nice, but for the proof of concept, it seemed more prudent to leverage the user's existing default editor.

ericrallen:
Closing this proposal as it was just a demo showing some possible approaches for future extensibility of the project.

sbendary25:
@ericrallen think i figured this out but might have botched the push to git. will tag u in my PR

6rz6:
**I can help with that**, 

Working on updating the local LLM docs with my latest changes and a full detailed tutorial. once done let me take a look and 
solve the issue, are, just verifying, before jumping in, are you working with Verbose on an using it on both open-interpreter and 
Litellm open side by side?  using break points also in Litellm?  and catching events?

✅i hope the answer is yes, 

because otherwise,even when you run only Op-In you clearly see the white output of Opin with the prefix and suffix to the user system prompt and then you see a green block of code, that is the forwarding of Litellm after phrasing a little the system message again before retrieving it back to Op-In.
 
❌If no and last sync was 5 months ago first better re-sync ♻ the branch 

I will submit a pull request soon for the docs and one get your feedback can sync pull newer files, and fix the intervention in the system prompt manually, in my case i need just to remove the word 'prompt' which is pushed to the system file among other chat related data and the LOGs.

If we are 5 months behind commits, no use investing in this version unless there is something i'm not aware of with main pull IMO better checkout / pull request the newest version and **solve** it with a '**man in the middle'**  as i did with integration 
of llamma.cpp :

image

@BellaBijl wdyt?

CC ->(@sbendary25 @ericrallen @Notnaton ) -> feedback would be highly welcome

PR #614, opened 2023-10-10
feat: add optional prompt token/cost estimate to %tokens

ericrallen:

This adds an optional `prompt` argument to the `%tokens` magic command that will estimate the tokens and cost of the provided prompt, which allows users to consider the implications of what they are going to send before it has an impact on their token usage.

OpenInterpreter-Tokens-Estimator

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [x] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/tests/test_interpreter.py
+++ b/tests/test_interpreter.py
@@ -1,5 +1,6 @@
 from random import randint
 import interpreter
+from interpreter.utils.count_tokens import count_tokens, count_messages_tokens
 import time
 
 interpreter.auto_run = True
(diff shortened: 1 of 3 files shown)


tests/test_interpreter.py
```diff
@@ -42,6 +43,28 @@ def test_reset():
     assert interpreter.messages == []
 
 
+def test_token_counter():
```
ericrallen, line 46: Added some basic tests to make sure the logic was sound.

We could also do some actual conversations with the model and then make sure that the tokens increment as expected when using the updated `interpreter.messages` array.

PR #676, opened 2023-10-22
Fix a crash in get_relevant_procedures_string

Notnaton:
Added a check if interpreter.procedures is None

It crashed on the for loop on my computer

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


KillianLucas:
Great catch @Notnaton. Actually, I'm not sure how it was working without this?

PR #680, opened 2023-10-23
Remove space from language

Notnaton:
The LLM might have put a space between code block start and language: "``` python ". Added Error handling to get better error messages.

Remove space from "language"
Added try/except to handle other errors with better error message.

Discord issue
https://discord.com/channels/1146610656779440188/1164496324591353946/1164496324591353946

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/core/respond.py
+++ b/interpreter/core/respond.py
@@ -2,6 +2,7 @@ from ..code_interpreters.create_code_interpreter import create_code_interpreter
 from ..utils.merge_deltas import merge_deltas
 from ..utils.display_markdown_message import display_markdown_message
 from ..utils.truncate_output import truncate_output
+from ..code_interpreters.language_map import language_map
 import traceback
 import litellm
 
@@ -113,12 +114,12 @@ def respond(interpreter):
 
                 # Get a code interpreter to run it
                 language = interpreter.messages[-1]["language"]
-                if language not in interpreter._code_interpreters:
-                    try:
+                if language in language_map:
+                    if language not in interpreter._code_interpreters:
                         interpreter._code_interpreters[language] = create_code_interpreter(language)
-                    except:
-                        print(f"Error: tried to create a code_interpreter with: '{language}'")
-                code_interpreter = interpreter._code_interpreters[language]
+                    code_interpreter = interpreter._code_interpreters[language]
+                else:
+                    print(f"{language} not in supported languagess.")
 
                 # Yield a message, such that the user can stop code execution if they want to
                 try:
--- a/interpreter/core/respond.py
+++ b/interpreter/core/respond.py
@@ -112,9 +112,12 @@ def respond(interpreter):
                     interpreter.messages[-1]["language"] = "shell"
 
                 # Get a code interpreter to run it
-                language = interpreter.messages[-1]["language"]
+                language = interpreter.messages[-1]["language"].replace(" ", "")
                 if language not in interpreter._code_interpreters:
-                    interpreter._code_interpreters[language] = create_code_interpreter(language)
+                    try:
+                        interpreter._code_interpreters[language] = create_code_interpreter(language)
+                    except:
+                        print(f"Error: tried to create a code_interpreter with: '{language}'")
                 code_interpreter = interpreter._code_interpreters[language]
 
                 # Yield a message, such that the user can stop code execution if they want to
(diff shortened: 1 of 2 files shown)


Notnaton:
Seems like the API key is missing?

interpreter/core/respond.py
```diff
@@ -112,9 +112,12 @@ def respond(interpreter):
                     interpreter.messages[-1]["language"] = "shell"
 
                 # Get a code interpreter to run it
-                language = interpreter.messages[-1]["language"]
+                language = interpreter.messages[-1]["language"].replace(" ", "")
```
ericrallen, line 115: I believe the [`language`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/llm/setup_openai_coding_llm.py#L127) property in the message object actually comes from the function call the language model makes to the [`execute`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/llm/setup_openai_coding_llm.py#L10) function, which only accepts one of the values in this list: `["python", "R", "shell", "applescript", "javascript", "html", "powershell"]`.

If you're looking to adjust how the markdown is generated and remove a potential space between the backticks and the language name that's causing an issue, I believe that would need to be handled in the `message` that comes back from the language model and the `.replace()` call here shouldn't actually have an effect.

interpreter/core/respond.py
```diff
@@ -112,9 +112,12 @@ def respond(interpreter):
...
                 if language not in interpreter._code_interpreters:
-                    interpreter._code_interpreters[language] = create_code_interpreter(language)
+                    try:
+                        interpreter._code_interpreters[language] = create_code_interpreter(language)
+                    except:
+                        print(f"Error: tried to create a code_interpreter with: '{language}'")
```
ericrallen, line 120: Should we consider checking against the available code interpreters in the [`language_map`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/code_interpreters/language_map.py#L10) rather than trying to create an interpreter that might fail?

interpreter/core/respond.py
```diff
@@ -113,9 +114,12 @@ def respond(interpreter):
...
                 language = interpreter.messages[-1]["language"]
-                if language not in interpreter._code_interpreters:
-                    interpreter._code_interpreters[language] = create_code_interpreter(language)
-                code_interpreter = interpreter._code_interpreters[language]
+                if language in language_map:
+                    if language not in interpreter._code_interpreters:
+                        interpreter._code_interpreters[language] = create_code_interpreter(language)
+                    code_interpreter = interpreter._code_interpreters[language]
+                else:
+                    print(f"{language} not in supported languages.")
```
Notnaton, line 122: What should happen when language is not in language_map?

ericrallen: Probably something like what you have here where we say something like: `Error: Open Interpreter does not currently support {language}.` and then we might need to yield that as output to make sure open interpreter recognizes the issue.

Notnaton:
Now it writes the code but disables the ability to run the code.
It also adds a error message on top.
bilde
@ericrallen

interpreter/core/respond.py
```diff
@@ -113,9 +114,21 @@ def respond(interpreter):
...
+                if language in language_map:
+                    if language not in interpreter._code_interpreters:
+                        interpreter._code_interpreters[language] = create_code_interpreter(language)
+                    code_interpreter = interpreter._code_interpreters[language]
+                else:
+                    #This still prints the code but don't allow code to run. Let's Open-Interpreter know through output message
+                    print(f"Error: Open-Interpreter do not currently support {language}.")
+
+                    interpreter.messages[-1]["output"] = ""
+                    output = "\n" + f"Error: Open-Interpreter do not currently support {language}."
```
ericrallen, line 126: Can we change these to the following message:

```python
error_output = f"Error: Open Interpreter does not currently support {language}."
```

And then use `error_output` in both places instead of repeating the string?

ericrallen:
@Notnaton Just had one tiny little change and then I think this is good to go.

This is a really nice improvement for the end user. Great work!

Notnaton:
@ericrallen Final fix done :)

ericrallen (approved):
Thanks for digging into the `language` edge case with the space and also making our handling of unsupported languages way more robust.

This is great.

KillianLucas:
Great work @Notnaton! Merging.

PR #682, opened 2023-10-23
Consider hiding the api-key

Notnaton:
Some people share screenshots with their API key visible.
Using getpass to hide the key should be considered

Added getpass to hide the key.

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/terminal_interface/validate_llm_settings.py
+++ b/interpreter/terminal_interface/validate_llm_settings.py
@@ -3,6 +3,7 @@ from ..utils.display_markdown_message import display_markdown_message
 import time
 import inquirer
 import litellm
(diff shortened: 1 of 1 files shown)


ericrallen (approved):
I think this is a great solution until we can move to something like [keyring](https://pypi.org/project/keyring/).

KillianLucas:
Agreed. `keyring` looks awesome for the future, and this solves the immediate problem of exposing API keys in screenshots.

Great work @Notnaton, merging now.

PR #683, opened 2023-10-23
Fixed issue with 3.5 Models hallucinating a function called "python", causing empty returns

Cobular:

I've found 3.5-turbo to be consistently hallucinating a function with the name "python", which causes an ugly UX of an empty return. Here's what it looks like:

```
{'role': 'assistant', 'content': None, 'function_call': <OpenAIObject at 0x11efe7c50> JSON: {
  "name": "python",
  "arguments": "import sys\nsys.executable"
}}
```
It turns out this is reasonably well known: https://community.openai.com/t/function-calling-a-python-function-is-frequently-called-even-though-it-does-not-exist-in-the-functions-parameter/264481/1, and that there's not a ton to be done here, other than embrace it. I've added the ability to just process that function, if it's called.

When working on raycast open-interpreter, I test with 3.5 turbo for cost reasons (I'm running the same question like 100 times to check UI stuff). I've found the issue to be problematic both on the command line and from the library form. 

I no longer get empty replies in any of the test cases I've found that once triggered this. My main test case is: `Check what python interpreter you have access to.`.

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/llm/setup_openai_coding_llm.py
+++ b/interpreter/llm/setup_openai_coding_llm.py
@@ -125,22 +125,21 @@ def setup_openai_coding_llm(interpreter):
                     arguments = parse_partial_json(arguments)
 
                     if arguments:
-                        if ("name" in arguments and arguments["name"] == "execute"):
-                            if (language is None
-                                and "language" in arguments
-                                and "code" in arguments # <- This ensures we're *finished* typing language, as opposed to partially done
-                                and arguments["language"]):
-                                language = arguments["language"]
-                                yield {"language": language}
-                            
-                            if language is not None and "code" in arguments:
-                                # Calculate the delta (new characters only)
-                                code_delta = arguments["code"][len(code):]
-                                # Update the code
-                                code = arguments["code"]
-                                # Yield the delta
-                                if code_delta:
-                                    yield {"code": code_delta}
+                        if (language is None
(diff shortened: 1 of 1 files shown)


ericrallen (approved):
This solves a very interesting issue with the `gpt-3.5-turbo` model, Thanks for digging in and working it out!

I do think it adds a lot of noise to an already noisy `debug_mode`, but the overall noisiness isn't the fault of this PR and the codebase doesn't provide an alternative.

Cobular:
Great! I'm happy to reduce the noise if you'd like - it's just the prints I was using to debug this on my end, but if needed I could pair them back to just a few critical ones

ericrallen:
Don't worry about the noise.

We'll need to address it at some point at a broader level in the codebase, but for now, it's still useful when trying to backtrack.

KillianLucas:
Wow. @Cobular this is like, a top 5 weird issue I figured we'd never track down. Brilliantly diagnosed and fixed. Merging now.

Cobular:
Sorry to poke this, but any idea when this'll get into a release? I just realized I was hitting it again on my new laptop which was running 0.1.10 instead of my fork.

PR #699, opened 2023-10-26
Proposal: Enforce consistent code style with black, isort, and pre-commit

ericrallen:

This adds `black`, `isort`, and `pre-commit` to the repository and runs `black` and `isort` on a `pre-commit` hook to help ensure that we have consistent code style across the codebase in order to eliminate a lot of the noise in Pull Requests where someone has accidentally formatted something differently due to their editor's settings.

- [ ] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,5 @@
 # Migrate code style to Black
 a463f15ffd9dcc36af4829b78b5e2a81d304f500
+
+# Rerun formatting after merge conflict resolution
+b6fac19ce47b0f16211948b907d8b3c55dcee985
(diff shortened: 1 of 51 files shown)


.git-blame-ignore-revs
ericrallen, line 1: This should preserve the regular git blame history for commits prior to the automated formatting commit and only show that commit as the git blame for newly added whitespace or removed whitespace changes.

ericrallen:
@KillianLucas the one thing I'm not sure about is if each developer needs to run `pre-commit install` locally, so it might be good to checkout this branch and see if the hooks run when you try to commit to it.

We can add a note to the Contributing Guide if necessary.

PR #701, opened 2023-10-26
Get more system info in debug mode

Notnaton:
Get some python, pip and os version and architecture.
This will help narrow down which systems and versions that don't work.

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


Notnaton:
There might be a better place to put it...

Notnaton:
Should also add cpu, gpu, LLM model, amount of ram...

ericrallen:
Thanks for the contribution, @Notnaton!

I really like the idea of having an easy way for folks to give us the relevant info easily.

I think it might be better as an optional parameter to the `%debug` magic command like `%debug info` that we can ask folks to run when triaging issues.

It would be great to have it output the info you have here as well as the Open Interpreter version and Model name.

If GPU and RAM info is easy to add, that would be nice for local issues, too.

This is a great idea.

interpreter/llm/setup_llm.py
```diff
@@ -5,13 +5,48 @@
...
+        os_name = platform.system()
+        os_release = platform.release()
+        os_version = platform.version()
+        print(f"OS Name: {os_name}")
+        print(f"OS Release: {os_release}")
+        print(f"OS Version: {os_version}")
+
+        # Print the architecture as well:
+        architecture = platform.architecture()
+        print(f"Architecture: {architecture[0]}")
```
ericrallen, line 48: I think it would be awesome to move all of this into [`interpreter/utils`](https://github.com/KillianLucas/open-interpreter/tree/main/interpreter/utils) as a `system_info` utility, or something like that, and then add a condition to the [`handle_debug()` magic command handler for `%debug`](https://github.com/KillianLucas/open-interpreter/blob/main/interpreter/terminal_interface/magic_commands.py#L66) to accept an `info` or `system` argument (not sure what the best name is) and then call this new utility so that folks can dynamically generate this debug info to copy and paste into Issues.

KillianLucas:
Nice addition @Notnaton! Yes, I agree with @ericrallen re: the placement. This seems like it should be in a util.

How about `%debug advanced`? This could be super noisy and print system info, then `%debug` is less noisy and doesn't print system info. Or maybe we just add this to `%debug` and slim down what gets printed through debug mode in general.

Also I've been thinking we should ask people to run `%debug` in the issue template, so more folks paste the output from that right off the bat.

Have also been thinking about a solution like `sentry`.

What do you both think?

Notnaton:
@KillianLucas yes it should be placed in utils. Would it be possible to run it in the last try/expect for open interpreter. So when it crashes it runs the code....
Apparently getting GPU info is annoyingly hard, without including more dependencies.

KillianLucas:
Oh wow, @Notnaton that's a great idea to run it after `except:`. If any error occurs we should run debug mode then raise the error.

I think GPU info isn't necessary at this stage— we should offload as much of those sorts of issues to other projects (perhaps LM Studio) anyway.

Notnaton:
Ok, will start on this tomorrow after work.
👍💤

Notnaton:
@ericrallen Should be ready now, I added the function to the "main" so when it is shut down it prints the info.

ericrallen (approved):
Thanks for another great addition, @Notnaton!

I think we’ll want to wire this up to a magic command, too, but we can do that in a separate PR.

KillianLucas:
Very clean! Great work @Notnaton. Merging now.

PR #703, opened 2023-10-26
fix: make all local models run through the html entity buffer

### LINKED ISSUE #697: Model generating HTML Entities instead of symbols in code
### Describe the bug

When I ask interpreter to create a python program, I get output like this:
```
input_dir = &#x27;/path/to/input/csv/files&#x27;                                                                                                 
output_dir = &#x27;/path/to/output/csv/files&#x27;  
```
This of course results in this error:
```
File "<unknown>", line 4                                                                                                                       
      input_dir = &#x27;/path/to/input/csv/files&#x27;                                                                                             
                  ^                                                                                                                                
  SyntaxError: invalid syntax  
```

### Reproduce

1. Ask interpreter to write a python program involving file IO
2. Wait

### Expected behavior

Correct use of ` ' ` character

### Screenshots

_No response_

### Open Interpreter version

0.1.10

### Python version

3.10.9

Ubuntu 23.10

### Additional context

_No response_

ericrallen:

This removes the mistral condition for running through the html entity buffer resolution.

Fixes #697

- [ ] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/llm/setup_local_text_llm.py
+++ b/interpreter/llm/setup_local_text_llm.py
@@ -91,13 +91,12 @@ def setup_local_text_llm(interpreter):
         buffer = ''  # Hold potential entity tokens and other characters.
 
         for token in ooba_llm.chat(messages):
-
-            if "mistral" not in repo_id.lower():
-                yield make_chunk(token)
-                continue
-
-            # For Mistral, we need to deal with weird HTML entities it likes to make.
-            # If it wants to make a quote, it will do &quot;, for example.
+            # Some models like to generate HTML Entities (like &quot;, &amp; &#x27;)
+            # instead of symbols in their code when used with Open Interpreter.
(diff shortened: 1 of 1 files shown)


ericrallen:
Based on [this comment](https://github.com/KillianLucas/open-interpreter/issues/697#issuecomment-1782448918) in #697.

I think this solves the issue with HTML entities.

We still need to investigate why hooking these models into Open Interpreter makes them prone to this behavior.

KillianLucas:
Very strange that it does. The system message is in markdown, and as you mentioned this isn't the default behavior of Mistral. Agreed this condition should be removed as I've noticed it in other local models too.

Issue #627, opened 2023-10-11
Compatibility Issue with Python 3.12 - Temporary Solution


TanmayDoesAI:
## Known Issue: Compatibility with Python 3.12

**Issue Description:**
As of the latest release, Open-Interpreter is facing compatibility issues with Python 3.12 due to certain dependencies, I believe `aiohttp`, not yet supporting this version. This may result in errors when attempting to run/install Open-Interpreter on Python 3.12.

**Temporary Solution:**
To work around this issue, it is recommended to create a separate Conda environment with a Python version lower than 3.12. You can use the following command to achieve this:

```bash
conda create -n open-interpreter-env python=3.11
conda activate open-interpreter-env
```

You can then proceed to use Open-Interpreter within this environment.

Dependency Update:
It's worth noting that the [aiohttp](https://github.com/aio-libs/aiohttp/) have released a beta version with the fix [here](https://github.com/aio-libs/aiohttp/issues/7685).

We appreciate your patience and understanding in this matter. If you encounter any further issues or have additional questions, please feel free to reach out to us on [Discord](https://discord.gg/6p3fD6rBVm) or by opening a new issue, providing all relevant details for expedited assistance.

Thank you for your continued support for Open-Interpreter!

bmwhite1611:
I tried downgrading to 3.11... but I have the new open-interpreter version... how do I downgrade to the last interpreter so at least it works?

MikeBirdTech:
3.12 does not play nice with tiktoken yet, and OI is dependent on it. Hopefully this changes soon! Will upgrade as soon as possible

JamesClarke7283:
> 3.12 does not play nice with tiktoken yet, and OI is dependent on it. Hopefully this changes soon! Will upgrade as soon as possible

Just some added context, i am sure everyone knows as it's closed.

I ran the unit test's today, and they seem to be fine for 3.12, with tiktoken 0.9.0.
https://github.com/openai/tiktoken/issues/264#issuecomment-2993468948

PR #704, opened 2023-10-26
chore: remove extraneous chroma package; clean up dependency list

ericrallen:

Based on #356 I tested to see if we could drop support down to Python `3.9`, but there were errors while trying to run the current version of interpreter - so I think we can close that one, and based on #627 and the many issues that pop up here and on Discord, we should explicitly state which versions are supported and indicate that `3.12` is not one of them at the moment.

Once all depencies support `3.12`, we can make the upper bound inclusive instead of exclusive.

`aiohttp` is an underlying dependency that is [still working on 3.12 support](https://github.com/aio-libs/aiohttp/issues/7675) and there may be others, too.

I believe the `chroma` package was installed in error as it isn't referenced anywhere, so I removed that and put the rest of the dependencies in alphabetical order (with the `python` version described first).

I also removed the commented out `pyreadline3` stuff to clean up the `pyproject.toml` as we can grab that from the git history if we need it again.

Closes #356

- [ ] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,33 +9,25 @@ authors = ["Killian Lucas <killian@openinterpreter.com>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.10"
-openai = "^0.28.0"
-rich = "^13.4.2"
-tiktoken = "^0.4.0"
-astor = "^0.8.1"
-git-python = "^1.0.3"
-tokentrim = "^0.1.9"
+python = ">=3.10,<3.12"
+
 appdirs = "^1.4.4"
-six = "^1.16.0"
-python-dotenv = "^1.0.0"
-
-inquirer = "^3.1.3"
-wget = "^3.2"
-huggingface-hub = "^0.17.3"
-litellm = "0.8.6"
-pyyaml = "^6.0.1"
-yaspin = "^3.0.1"
-ooba = "^0.0.21"
-chroma = "^0.2.0"
+astor = "^0.8.1"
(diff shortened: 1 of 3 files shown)


Adityaranjan932:
Compatibility hiccups with Python updates are pretty common. Have you checked if there's an updated version of aiohttp or any plans for Python 3.12 support from their end?

ericrallen:
@Adityaranjan932 I linked to the Issue where they are tracking support for 3.12 in the PR description.

pyproject.toml
```diff
@@ -9,33 +9,25 @@ authors = ["Killian Lucas <killian@openinterpreter.com>"]
...
+python-dotenv = "^1.0.0"
 pyyaml = "^6.0.1"
+rich = "^13.4.2"
+six = "^1.16.0"
+tiktoken = "^0.4.0"
+tokentrim = "^0.1.9"
+wget = "^3.2"
 yaspin = "^3.0.1"
-ooba = "^0.0.21"
-chroma = "^0.2.0"
```
ericrallen, line 30: This is the actual package we’re removing.

I think it might have been `poetry add`ed by accident as we don’t actually seem robust it for anything and `chromadb` - also installed - is the package for the Chroma Vector Database.

KillianLucas: Yes, my mistake— I must have thought that was `chromadb`.

PR #709, opened 2023-10-27
chore: add better docs for safe_mode; load semgrep if available

ericrallen:

Update docs for `safe_mode`, dynamically load `semgrep` if it's available.

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [x] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


docs/SAFE_MODE.md
```diff
@@ -0,0 +1,61 @@
...
+
+**⚠️ Safe mode is experimental and does not provide any guarantees of safety or security.**
+
+Open Interpreter is working on providing an experimental safety toolkit to help you feel more confident running the code generated by Open Interpreter.
+
+
+Install Open Interpreter with the safety toolkit dependencies as part of the bundle:
+
+```shell
+pip install open-interpreter[safe]
```
ericrallen, line 11: We can leverage the `tool.poetry.extras` configuration to allow folks to optionally install `semgrep` and other safety tools via the `[safe]` extra.

interpreter/terminal_interface/terminal_interface.py
```diff
@@ -17,8 +18,9 @@ def terminal_interface(interpreter, message):
             "**Open Interpreter** will require approval before running code."
         ]
 
-        if interpreter.safe_mode != "off":
-            interpreter_intro_message.append(f"**Safe Mode**: {interpreter.safe_mode}\n\n>Note: **Safe Mode** requires `semgrep` (`pip install semgrep`)")
+        if interpreter.safe_mode == "ask" or interpreter.safe_mode == "auto":
+            if not check_for_package("semgrep"):
+                interpreter_intro_message.append(f"**Safe Mode**: {interpreter.safe_mode}\n\n>Note: **Safe Mode** requires `semgrep` (`pip install semgrep`)")
```
ericrallen, line 23: Now we only need to show the `semgrep` message if it isn't already in the user's current environment. If `semgrep` is available and `safe_mode` is enabled, `semgrep` will be automatically loaded.

docs/SAFE_MODE.md
KillianLucas: YES. I love this idea. 

I think it would be great to make a slimmed down version of OI called `[core]` (wouldn't even include `rich` or any terminal interface), `[safe]` for semgrep, perhaps docker, etc, then eventually `[all]` if we build features beyond safe mode with dependencies.

KillianLucas:
Great work Eric. Really happy to start seeing [extras] in OI— opens the door to OI serving different kinds of users more precisely.

PR #741, opened 2023-11-09
Continue to run despite unsupported language

Notnaton:
Tries to not stop interpreter when using unsupported language. 
Still not perfect, looking for testers and feedback. 

https://discord.com/channels/1146610656779440188/1171921044496470106

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [x] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/terminal_interface/terminal_interface.py
+++ b/interpreter/terminal_interface/terminal_interface.py
@@ -151,12 +151,16 @@ def terminal_interface(interpreter, message):
 
                             scan_code(code, language, interpreter)
 
+                        interpreter.run_code_promt = False
                         response = input(
                             "  Would you like to run this code? (y/n)\n\n  "
                         )
                         print("")  # <- Aesthetic choice
 
                         if response.strip().lower() == "y":
+                            interpreter.run_code_promt = True
+
+                        if interpreter.run_code_promt:
                             # Create a new, identical block where the code will actually be run
                             # Conveniently, the chunk includes everything we need to do this:
                             active_block = CodeBlock()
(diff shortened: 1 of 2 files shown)


interpreter/terminal_interface/terminal_interface.py
```diff
@@ -151,12 +151,16 @@ def terminal_interface(interpreter, message):
...
+                        interpreter.run_code_promt = False
                         response = input(
                             "  Would you like to run this code? (y/n)\n\n  "
                         )
                         print("")  # <- Aesthetic choice
 
                         if response.strip().lower() == "y":
+                            interpreter.run_code_promt = True
+
+                        if interpreter.run_code_promt:
```
ericrallen, line 163: I think this is an unnecessary condition.

We can just set the property to true and follow the code path in this condition without needing to check the value of the property we just set.

Notnaton: There is a third one in response.py

ericrallen: Yeah, for that we need to set the `interpreter.run_code_promt` (which I just noticed has a typo, can we make that `interpreter.run_code_prompt`), but we don’t need this `if` statement because at this point in the code that property can only have one value because we just set it a couple of lines above.

Notnaton: I added it to stop it running the code regardless of y/n. It just didn't output the result when saying n. 
I'll double check and come back to this after work 👍

Notnaton:
See:
https://github.com/KillianLucas/open-interpreter/pull/745

Much better solution than my code

PR #742, opened 2023-11-09
chore: popen using utf8 encoding

### LINKED ISSUE #209: UnicodeDecodeError - help will be appriciate!
_Exception in thread Thread-1 (save_and_display_stream):
  File "C:\Users\ziv\AppData\Local\Programs\Python\Python311\Lib\threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "C:\Users\ziv\AppData\Local\Programs\Python\Python311\Lib\threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\ziv\AppData\Local\Programs\Python\Python311\Lib\site-packages\interpreter\code_interpreter.py", line
293, in save_and_display_stream
    for line in iter(stream.readline, ''):
  File "C:\Users\ziv\AppData\Local\Programs\Python\Python311\Lib\encodings\cp1255.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8e in position 3284: character maps to <undefined>_

I am a Windows User, running with Conda, on Python version 3.11.2
 should I change the encoding?

fatelei:

fix #209  using utf8 encoding instead of cp1255

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/code_interpreters/subprocess_code_interpreter.py
+++ b/interpreter/code_interpreters/subprocess_code_interpreter.py
@@ -1,3 +1,4 @@
+import os
 import queue
 import subprocess
 import threading
@@ -40,6 +41,8 @@ class SubprocessCodeInterpreter(BaseCodeInterpreter):
         if self.process:
             self.terminate()
 
+        my_env = os.environ.copy()
+        my_env['PYTHONIOENCODING'] = 'utf-8'
         self.process = subprocess.Popen(
             self.start_cmd.split(),
             stdin=subprocess.PIPE,
@@ -48,6 +51,7 @@ class SubprocessCodeInterpreter(BaseCodeInterpreter):
             text=True,
             bufsize=0,
             universal_newlines=True,
+            env=my_env
         )
         threading.Thread(
             target=self.handle_stream_output,


interpreter/code_interpreters/subprocess_code_interpreter.py
```diff
@@ -40,6 +40,7 @@ def start_process(self):
         if self.process:
             self.terminate()
 
+        my_env['PYTHONIOENCODING'] = 'utf-8'
```
ericrallen, line 43: Should this be `os.environ['PYTHONIOENCODING'] = 'utf-8'`?

I don't think the `my_env` variable exists.

fatelei: https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING

ericrallen: The `PYTHONIOENCODING` part looks good, but I believe the `my_env` variable doesn’t exist or refer to anything.

fatelei: ok

interpreter/code_interpreters/subprocess_code_interpreter.py
```diff
@@ -40,6 +41,8 @@ def start_process(self):
         if self.process:
             self.terminate()
 
+        my_env = os.environ.copy()
+        my_env['PYTHONIOENCODING'] = 'utf-8'
```
ericrallen, line 45: Is there a reason to add this indirection via a copy of the environment object?

Can’t we just set the property on `os.environ` directly?

fatelei:
if we want to customize env not affect system env, copy it may be better

ericrallen (approved):
Based on my limited testing, this seems to work as expected.

I've never encountered the error that other folks were getting, though.

fatelei:
> Based on my limited testing, this seems to work as expected.
> 
> I've never encountered the error that other folks were getting, though.

what about unit tests

ericrallen:
@fatelei If you can think of a good way to recreate a situation that triggers the error and then add a unit test where we make sure that `interpreter` can still execute code, that would be great.

I haven't had a chance to put brainpower towards it because I'm not sure what the underlying environment setting that causes an issue in the first place is.

I guess we could just set `PYTHONIOENCODING` to something else in the test, but then we'd just be testing if setting the encoding works rather than if setting it and passing the environment to the subprocess fixes the underlying issue with the system having a different encoding.

KillianLucas:
This seems like a good idea— `utf-8` handles diverse output chars including the European chars the issues seemed to be having difficulty with. Pushing to pip soon so I'm going to merge this.

I think in the future a good unit test would try out a bunch of different non-English chars, but something tells me that might only be useful once we can be testing things across a bunch of different OS/terminals, not just the github ubuntu non-GUI one.

Great work @fatelei and thanks @ericrallen for reviewing!

PR #745, opened 2023-11-10
Unsupported lang fix

CyanideByte:
When unsupported language is encountered, it continues the response so the model can try workarounds and continue with the remaining steps instead of immediately stopping. Also added a small check so that if the model tries to read out the code a 2nd time it doesnt loop.

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [x] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/core/respond.py
+++ b/interpreter/core/respond.py
@@ -132,7 +132,7 @@ If LM Studio's local server is running, please try a language model with a diffe
                     interpreter.messages[-1]["language"] = "shell"
 
                 # Get a code interpreter to run it
-                language = interpreter.messages[-1]["language"].lower()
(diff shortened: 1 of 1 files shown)


Notnaton:
@ericrallen I closed my PR in favor of this one. It is much simpler and doesn't require my hacky solution :)
I have tested this PR, works as intended on my machine.
Great work @CyanideByte

KillianLucas:
Very clean patch. Thanks @CyanideByte for fixing this! And thanks @Notnaton for your PR and for testing this one. Merging.

PR #767, opened 2023-11-16
remove large files from git history

### LINKED ISSUE #214: Large files in git history
**Describe the bug**
Cloning down the repo takes ~89MB due to large files that were committed in the past and removed from the current code base, but not from the git history.

**To Reproduce**
`du -hs .git`

**Expected behavior**
The repo should be under 2MB when the unnecessary large files (venv folder and .whl files) are removed from git history

**Additional context**
[BFG](https://rtyley.github.io/bfg-repo-cleaner/) can clean them up, though it will involve rewriting git history and require a rebase from all open PRs. My view is it is better to wear the pain once while the project is still young than just ignore it.

You could actually do this with open interpreter if you wanted to - here is one way of doing it https://youtu.be/SIVDUDfKcJo?si=4COYdCxb8_CAaclP&t=425

joshuavial:
### Reference any relevant issue (Fixes #214)

- [x ] I have performed a self-review of my code:

```
git remote add joshuavial https://github.com/joshuavial/open-interpreter
git diff origin/main joshuavial/main
```


KillianLucas:
Extraordinary work @joshuavial.

This really helps the repo in the long term, and it was a huge chore I know. Merged, and it seems recent PRs can even effortlessly be merged (I have no idea how).

PR #770, opened 2023-11-17
Ask user if they want to switch models when they can't access gpt-4. Turbo models updated.

CyanideByte:
Asks user if they want to switch to 3.5-turbo if they don't have access to gpt-4.
Detects old model names and changes them to the new turbo models.

https://discord.com/channels/1146610656779440188/1175102056160301147

- [x] Windows
- [ ] MacOS
- [x] Linux

- [x] GPT4
- [x] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


KillianLucas:
Great work @CyanideByte! Tons of people get a fresh API key for OI, and it's very confusing when it breaks on `gpt-4`. Super helpful addition. Also love setting `max_tokens` and `context_window` on the new models. Merging now.

interpreter/terminal_interface/start_terminal_interface.py
```diff
@@ -283,6 +283,23 @@ def start_terminal_interface(interpreter):
         # Doesn't matter
         pass
 
+    # At some point in the future these model names redirects wont be necessary anymore, but legacy names will remain for a while
+    if interpreter.model == "gpt-4" or interpreter.model == "gpt-4-32k":
+        interpreter.model = "gpt-4-1106-preview"
```
KillianLucas, line 288: (btw going to remove these lines after merging, as I've found `gpt-4-1106-preview` to be worse in some ways and we should let folks choose!)

PR #790, opened 2023-11-24
Better config 

Notnaton:

- [ ] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -66,7 +66,16 @@ class Interpreter:
             print(f"Extending configuration from `{config_path}`")
 
         config = get_config(config_path)
-        self.__dict__.update(config)
+        
+        # Accessing the default profile parameter
+        default_profile = config.get('default_profile', None)
+        if default_profile is not None:
+            print(f"default_profile: {default_profile}")
+        else:
+            print("default_profile is not set in the configuration")
+
+        self.__dict__.update(config['base']) #Get system messages
+        self.__dict__.update(config[default_profile])
 
     def chat(self, message=None, display=True, stream=False):
         if stream:
--- a/interpreter/terminal_interface/config.yaml
+++ b/interpreter/terminal_interface/config.yaml
@@ -1,14 +1,81 @@
-system_message: |
-  You are Open Interpreter, a world-class programmer that can complete any goal by executing code.
-  First, write a plan. **Always recap the plan between each code block** (you have extreme short-term memory loss, so you need to recap the plan between each message block to retain it).
-  When you execute code, it will be executed **on the user's machine**. The user has given you **full and complete permission** to execute any code necessary to complete the task.
-  If you want to send data between programming languages, save the data to a txt or json.
-  You can access the internet. Run **any code** to achieve the goal, and if at first you don't succeed, try again and again.
-  You can install new packages.
-  When a user refers to a filename, they're likely referring to an existing file in the directory you're currently executing code in.
-  Write messages to the user in Markdown.
-  In general, try to **make plans** with as few steps as possible. As for actually executing code to carry out that plan, for *stateful* languages (like python, javascript, shell, but NOT for html which starts from 0 every time) **it's critical not to try to do everything in one code block.** You should try something, print information about it, then continue from there in tiny, informed steps. You will never get it on the first try, and attempting it in one go will often lead to errors you cant see.
-  You are capable of **any** task.
-local: false
-model: "gpt-4"
-temperature: 0
+#Choose which profile to run using "interpreter"
+#use interpreter --profile <profile> to override
+#interpreter --fast / --local switches the profile too.
+default_profile: "local"
+
+openai-vision:
+  model: "openai/"
+  max_tokens: 1024
+  context_window: 3000
+
+gpt-4:
+  model: "openai/gpt-4"
+  max_tokens: 1024
+  context_window: 3000
+
+gpt-3:
+  model: "openai/gpt-3.5-turbo"
+  max_tokens: 1024
+  context_window: 3000
+
+local:
+  model: "openai/local"
+  api_base: "http://localhost:1234/v1"
+  max_tokens: 1024
+  context_window: 3000
+  api_key: "0"
+
+base:
+  temperature: 0
+  system_message: |
+    You are Open Interpreter, a world-class programmer that can complete any goal by executing code.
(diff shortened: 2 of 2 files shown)


Notnaton:
Tested Local config

interpreter/terminal_interface/start_terminal_interface.py
```diff
@@ -259,47 +233,28 @@ def start_terminal_interface(interpreter):
...
-        interpreter.context_window = 110000
-        interpreter.max_tokens = 4096
+        interpreter.load_config("vision")
 
-        display_markdown_message("> `Vision` enabled **(experimental)**\n")
+    if args.local:
+        interpreter.load_config("local")
+
+    if args.profile:
+        interpreter.load_config(args.profile)
```
Notnaton, line 245: This should happen before the other args. so --model overwrites the config

interpreter/terminal_interface/start_terminal_interface.py
```diff
@@ -238,7 +212,7 @@ def start_terminal_interface(interpreter):
             if attr_name == "config_file":
                 user_config = get_config_path(attr_value)
                 interpreter.config_file = user_config
-                interpreter.extend_config(config_path=user_config)
+                interpreter.load_config()
             else:
                 setattr(interpreter, attr_name, attr_value)
```
Notnaton, line 217: This needs testing

interpreter/core/core.py
```diff
@@ -56,17 +58,37 @@ def __init__(self):
...
+
+        # Use default_profile if no specific profile is provided
+        if profile is None:
+            profile = config.get('default_profile', None)
+
+        # Load profile-specific configuration
+        if profile in config:
+            #Load default values first
+            self.update_attributes(config['base'])
+            self.update_attributes(config[profile])
```
Notnaton, line 78: doesn't append system_message

interpreter/core/core.py
```diff
@@ -56,17 +58,37 @@ def __init__(self):
...
+        else:
+            print(f"Profile '{profile}' not found in the configuration file.")
+            raise
+    
+    def update_attributes(self, config):
```
leiftaylorcb, line 83: While these methodology of extracting self attributes from the config gives the appearance of DRYer, cleaner code, I think it introduces some issues that outweigh the cleanliness.
If we allow users to directly modify the "self" attributes of the class we potentially expose:

**Maintainability Challenges:** Such direct modifications can complicate code maintenance and updates, risking compatibility issues with existing configurations.

**Validation Concerns:** This method may bypass essential validation checks, possibly leading to inconsistencies or invalid program states.

**Encapsulation Breach:** It goes against the OOP principle of encapsulation, crucial for safeguarding an object's internal state and exposing it only through a defined interface.

**Security Risks:** This could introduce vulnerabilities, particularly with critical functionality attributes, allowing potential unauthorized access or manipulation.

This could be considered an anti-pattern.
The simplest way to handle this would be to use setters and simply have the yaml read something like:
```
base:
  # Settings
  local: False
  auto_run: False
  debug_mode: False
  max_output: 2000
  safe_mode: "off"
  disable_procedures: False
  launch_message: ""
  ```
  
  and then assign them explicity in the set_attributes method like:
  ```
  def set_attributes(self, config):
    if 'local' in config:
        self.set_local(config['local'])
    if 'auto_run' in config:
        self.set_auto_run(config['auto_run'])
    if 'debug_mode' in config:
        self.set_debug_mode(config['debug_mode'])
```

The issue here is that if we have like 80+ attributes, this will get cumbersome.
So the best practice in this case would be to use schema based validation, e.g.

Define a schema model with pydantic:
```
from pydantic import BaseModel

class ConfigSchema(BaseModel):
    local: bool = False
    auto_run: bool = False
    debug_mode: bool = False
    max_output: int = 2000
    safe_mode: str = "off"
    disable_procedures: bool = False
    launch_message: str = ""
    # Define other settings as needed...
    # ...

    # Optionally, include nested configurations or lists if required
```

Update the class to apply the config:
```
import yaml
from pydantic import ValidationError

def load_and_validate_config(file_path: str) -> ConfigSchema:
    with open(file_path, 'r') as file:
        config_data = yaml.safe_load(file)

    try:
        return ConfigSchema(**config_data)
    except ValidationError as e:
        print("Validation Error:", e)
        raise
```

Then apply the config:
```
import yaml
from pydantic import ValidationError

class Interpreter:
    # ... other methods and the __init__ method ...

    def load_config(self, profile=None):
        with open(self.config_file, 'r') as file:
            config_data = yaml.safe_load(file)

        # Use the default profile if no specific profile is provided
        if profile is None:
            profile = config_data.get('default_profile', None)

        try:
            # Validate and load the base configuration
            base_config = ConfigSchema(**config_data.get('base', {}))
            self.apply_config(base_config)

            # Validate and load the profile-specific configuration
            if profile in config_data:
                profile_config = ConfigSchema(**config_data[profile])
                self.apply_config(profile_config)
            else:
                print(f"Profile '{profile}' not found in the configuration file.")
                raise ValueError(f"Profile '{profile}' not found.")
        except ValidationError as e:
            print("Configuration validation error:", e)
            raise

    def apply_config(self, config: ConfigSchema):
        # Apply configuration to the class attributes
        self.local = config.local
        self.auto_run = config.auto_run
        self.debug_mode = config.debug_mode
        self.max_output = config.max_output
        self.safe_mode = config.safe_mode
        self.disable_procedures = config.disable_procedures
        self.launch_message = config.launch_message
        # Apply other settings as needed...
        # ...

```

KillianLucas:
**Extraordinary** work @Notnaton. Profiles are a fantastic, future-proof idea that removes much of the complexity of "vision" and "local", which never should have been simple booleans! 

And many thanks for reviewing it @leiftaylorcb. Agreed that some validation would be a good idea, as long as we can keep it as simple as possible— ideally that schema would be derived from the dictionary at the top of `start_terminal_interface.py`, or we would replace that dictionary with a schema somehow.

In the next few days I'm going to merge this then continue working on it: 

Something I should have mentioned was that the [roadmap](https://github.com/KillianLucas/open-interpreter/blob/main/docs/ROADMAP.md) has us moving all `config` stuff into the `/terminal_interface`, and entirely out of the `/core` python package (which IMO should be much more like a standard, lightweight python package, and less like an app with state between sessions— that's the `terminal_interface`'s job).

If you'd like to start taking in this direction that would be helpful, but don't feel obliged too (as I'll re-implement it in this way after I merge it).

PR #803, opened 2023-12-01
Fix %tokens command crash

CyanideByte:
- Fixed %tokens command where count_tokens fails to find tokenizer because of openai/ in the model name
- If tokenizer for a model isn't found, defaults to gpt-4 tokenizer

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [x] GPT4
- [x] GPT3
- [x] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


KillianLucas:
Clean fix! Thanks @CyanideByte. Really like that it gives an estimation if it can't find it. Merging.

PR #807, opened 2023-12-03
Update README.md

Notnaton:
Added documentation for using local models using python

https://discord.com/channels/1146610656779440188/1180576022635237417

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [x] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,19 @@ interpreter.chat("Plot AAPL and META's normalized stock prices") # Executes a si
 interpreter.chat() # Starts an interactive chat
 ```
 
+#### Local models
+
+```python
+import interpreter
+
+interpreter.local = True #does nothing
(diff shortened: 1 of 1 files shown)


Notnaton:
Just a general warning
```
import interpreter

interpreter.local = True
interpreter.chat()
```
This uses OpenAI API if OPENAI_API_KEY is set.

KillianLucas:
> #does nothing

lmao this is why I think ".offline" makes much more sense, so it's not conflated with the terminal interface's `--local` — right now that bool disables _some_ online features like [Open Procedures](https://open-procedures.replit.app/). Merging and will add that comment, I agree this is frustratingly unintuitive. Thank you so much for adding clarifying docs.

PR #808, opened 2023-12-03
Inserts dummy api key if missing

CyanideByte:
Inserts dummy api key if missing (fixes python scripts accessing self hosted solutions crashing without an api key)

https://discord.com/channels/1146610656779440188/1180576022635237417

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [x] GPT4
- [x] GPT3
- [x] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B


KillianLucas:
Thanks for fixing this @CyanideByte! I'll merge but will play with fixing this closer to the problem for better maintainability. I feel like this is a problem with LiteLLM, so the solution should probably insert a dummy key right before we send it to LiteLLM.

Issue #794, opened 2023-11-27
If config is empty interpreter crashes


Notnaton:
### Describe the bug

Deleting the contents of the config file crashes interpreter

If there is someone who wants to try to contribute to Open Source Software for the first time, here is a good issue to work on. 
@ me if you need assistance, I'll help you make your first PR :teacher: 
You'll need some basic understanding of python :snake: 

### Reproduce

```
interpreter --config
```
delete content, save and close.
```
interpreter --config
```

### Expected behavior

Open the config file

### Screenshots

![image](https://github.com/KillianLucas/open-interpreter/assets/4324290/291d6e41-3a7a-45c2-ba74-3be6baf614c6)

### Open Interpreter version

0.1.16

### Python version

3.11.6

### Operating System name and version

Windows 11

Vikk-17:
Hey @Notnaton ! I want to this. Please help me by providing details what to do and how to do. :)

PR #811, opened 2023-12-03
 return empty dict if config file is empty

sbendary25:
`yaml.safe_load` returns `None` from the `get_config` function if the config file is empty. This change checks for `None` value and instead returns `{}` so `extend_config` call doesn't throw exception when config file is empty.

https://github.com/KillianLucas/open-interpreter/issues/794

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/terminal_interface/utils/get_config.py
+++ b/interpreter/terminal_interface/utils/get_config.py
@@ -47,9 +47,18 @@ def get_config_path(path=user_config_path):
 
 def get_config(path=user_config_path):
     path = get_config_path(path)
+    
+    config = None
 
     with open(path, "r") as file:
         config = yaml.safe_load(file)
-        if config is None:
-            config = {}
-        return config
+        if not config is None:
+            return config
+
+    if config is None:
+        # Deleting empty file because get_config_path copies the default if file is missing
+        os.remove(path)
+        path = get_config_path(path)
+        with open(path, "r") as file:
+            config = yaml.safe_load(file)
+            return config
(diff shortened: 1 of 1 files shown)


CyanideByte (changes requested):
I tested the code @sbendary25 provided and it works great. I think you could go a step further and have it repair the config file using the default one. Here is some tested code below that does that.
```
def get_config(path=user_config_path):
    path = get_config_path(path)
    
    config = None

    with open(path, "r") as file:
        config = yaml.safe_load(file)
        if not config is None:
            return config

    if config is None:
        os.remove(path)
        path = get_config_path(path)
        with open(path, "r") as file:
            config = yaml.safe_load(file)
            return config
```

sbendary25:
added @CyanideByte suggestion [a90cc8f](https://github.com/KillianLucas/open-interpreter/pull/811/commits/a90cc8fa0db14a675e2558e57cb3d63d965c7b60)

interpreter/terminal_interface/utils/get_config.py
```diff
@@ -47,9 +47,18 @@ def get_config_path(path=user_config_path):
...
     path = get_config_path(path)
+    
+    config = None
 
     with open(path, "r") as file:
         config = yaml.safe_load(file)
-        if config is None:
-            config = {}
-        return config
+        if not config is None:
```
Notnaton, line 55: `if config is not None:` is better syntax
image
from: https://pep8.org/

sbendary25:
@Notnaton also added a message to inform user that default config is being applied

sbendary25:
@Notnaton bumping this to be closed and merged.

KillianLucas:
Great work @sbendary25. And thanks @CyanideByte and @Notnaton for your suggestions and reviews. Merging now.

(also thanks for your comments on the latest `--os` updates Samer— you have a great attention to detail)

PR #830, opened 2023-12-12
Feature/edit code

sbendary25:
scan and edit recently edited code by creating a while loop around the "executing" chunk conditional block 
this is based on the following PR by [ericrallen](https://github.com/ericrallen)
https://github.com/KillianLucas/open-interpreter/pull/612

- [ ] Windows
- [x] MacOS
- [ ] Linux

- [x] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B

### DIFF
--- a/interpreter/cli/cli.py
+++ b/interpreter/cli/cli.py
@@ -1,10 +1,8 @@
 import argparse
-import subprocess
 import os
-import platform
 import pkg_resources
 import appdirs
-from ..utils.display_markdown_message import display_markdown_message
+from ..utils.open_file import open_file
 from ..terminal_interface.conversation_navigator import conversation_navigator
 
 arguments = [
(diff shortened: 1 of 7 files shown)


KillianLucas:
Nice, thanks @sbendary25 for working on this and of course @ericrallen for the incredible initial PR. 

If this can be adapted to the current codebase (I think this PR is from a few weeks ago) I'd be happy to merge. I'd probably simplify it tho — for example I think this shouldn't interface with any code scanning utils (any code the user has written = deemed safe by them, I think). Is the only reason for that crossover because we also use a tempfile to scan code?

I'd probably want to reduce it to a magic command which edits the last code block, then continues OI from there.

This way user's can:
- Decline to run a code block by pressing `N`
- Press `%e` or `%edit` or something
- We create a temporary file, populate it with the last code block
- ↑ Open it in the user's preferred editor
- `yaspin` fires up, "Code editor opened. Save your changes then press `ENTER` here to run modified code."
- They edit the file, save it, press `ENTER`
- We edit the last code block in `interpreter.messages`, re-display it, continue OI from there

Kreijstal:
I am excited for this one

Eliaskhal:
I think what we can do:
1- ask the user to set their favorite editor once (maybe in the beginning), for example 'code'.
2- Instead of asking the user if they want to run the code or no, we can add another option to run the code.
3- if they choose to edit the code: save the code in a temp file, then run a subprocess  opening the editor and waiting for it to close.
```py
if system_platform == 'windows':
        subprocess.run(['start', 'your_editor', filename], shell=True)
```
When the user is done we can give the option to re-edit the code or run it.

Notnaton:
https://github.com/KillianLucas/open-interpreter/pull/942

PR #840, opened 2023-12-19
Fixed the error when using Azure OpenAI API

wsbao:
Run the interpreter like this:

```console
interpreter --model azure/{engine} --api_base {AZURE_API_BASE} --api_key {AZURE_API_KEY} --api_version {AZURE_API_VERSION}
```

- [x] Windows
- [ ] MacOS
- [ ] Linux

- [ ] GPT4
- [ ] GPT3
- [ ] Llama 7B
- [ ] Llama 13B
- [ ] Llama 34B
- [x] GPT4 on Azure

### DIFF
--- a/interpreter/core/llm/utils/convert_to_openai_messages.py
+++ b/interpreter/core/llm/utils/convert_to_openai_messages.py
@@ -41,6 +41,9 @@ def convert_to_openai_messages(
                         "code": message["content"],
                     },
                 }
+                # Add empty content to avoid error "openai.error.InvalidRequestError: 'content' is a required property - 'messages.*'"
+                # especially for the OpenAI service hosted on Azure
+                new_message["content"] = "" 
             else:
                 new_message[
                     "content"
--- a/interpreter/terminal_interface/start_terminal_interface.py
+++ b/interpreter/terminal_interface/start_terminal_interface.py
@@ -120,6 +120,13 @@ def start_terminal_interface(interpreter):
             "attribute": {"object": interpreter.llm, "attr_name": "api_key"},
         },
         {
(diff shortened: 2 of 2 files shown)


KillianLucas:
Thank you so much @wsbao for building this before we push to `pip`. Sure to save tons of Azure users a headache. Merging now.

wsbao:
Using the up to date code, the command line interface does not work properly. This might be caused by the change of other code. I also notice that: when I run the interpreter in the command line using debug_mode, the model shows "openai/azure/{engine}" where the "openai" prefix is added.

# Interpreter Info

        Vision: False
        Model: openai/azure/{engine}
        Function calling: None
        Context window: None
        Max tokens: None

        Auto run: False
        API base: https://***.openai.azure.com
        Offline: False

wsbao:
I think the problem arises from the following code snippet in "start_terminal_interface.py" from line 560 to 564
```python
    # If we've set a custom api base, we want it to be sent in an openai compatible way.
    # So we need to tell LiteLLM to do this by changing the model name:
    if interpreter.llm.api_base:
        if not interpreter.llm.model.lower().startswith("openai/"):
            interpreter.llm.model = "openai/" + interpreter.llm.model
```
After I changed it like this, it works again:
```python
    # If we've set a custom api base, we want it to be sent in an openai compatible way.
    # So we need to tell LiteLLM to do this by changing the model name:
    if interpreter.llm.api_base:
        if not interpreter.llm.model.lower().startswith("openai/") and \
           not interpreter.llm.model.lower().startswith("azure/"):
            interpreter.llm.model = "openai/" + interpreter.llm.model
```

PR #884, opened 2024-01-08
Fix cli crash during screenshot generation

### LINKED ISSUE #878: OSError: cannot write mode RGBA as JPEG
### Describe the bug

CleanShot 2024-01-06 at 20 17 51@2x

I get this error when trying to run interpreter in --os mode. 

### Reproduce

---------------------------------------------------------------------------
  KeyError                                  Traceback (most recent call last)
  File ~/miniconda3/lib/python3.10/site-packages/PIL/JpegImagePlugin.py:639,
  in _save(im, fp, filename)
      638 try:
  --> 639     rawmode = RAWMODE
      640 except KeyError as e:

  KeyError: 'RGBA'

  The above exception was the direct cause of the following exception:

  OSError                                   Traceback (most recent call last)
  File ~/miniconda3/lib/python3.10/site-packages/PIL/Image.py:643, in
  Image._repr_image(self, image_format, **kwargs)
      642 try:
  --> 643     self.save(b, image_format, **kwargs)
      644 except Exception as e:

  File ~/miniconda3/lib/python3.10/site-packages/PIL/Image.py:2413, in
  Image.save(self, fp, format, **params)
     2412 try:
  -> 2413     save_handler(self, fp, filename)
     2414 except Exception:

  File ~/miniconda3/lib/python3.10/site-packages/PIL/JpegImagePlugin.py:642,
  in _save(im, fp, filename)
      641     msg = f"cannot write mode {im.mode} as JPEG"
  --> 642     raise OSError(msg) from e
      644 info = im.encoderinfo

  OSError: cannot write mode RGBA as JPEG

  The above exception was the direct cause of the following exception:

  ValueError                                Traceback (most recent call last)
  File
  ~/miniconda3/lib/python3.10/site-packages/IPython/core/formatters.py:344, in

### Expected behavior

expecting to be able to continue sending correctly the image of the screen

### Screenshots

_No response_

### Open Interpreter version

 0.2.0

### Python version

3.10.9

MacOS Sonoma 14.2.1

### Additional context

_No response_

double-thinker:
### Description of changes:
Implemented a fix that addresses an RGBA handling issue during screenshot generation when using IPython with auto-display.

The issue happens because `show=True` is set in `Display.snapshot`, and the screenshot generated on a Mac has an alpha channel (RGBA instead of RGB). When IPython attempts to display this image and convert it to JPEG, the process crashes (the exact cause is unclear).

This PR introduces a simple fix that enforces RGB mode in all cases.

I chose not to alter the default value of `show` because there may be use cases that depend on the current behavior, and this change is less likely to introduce new issues.

In the following commit, I performed a minor refactor related to the change I made: instead of saving/loading the image into a file that is neither used nor exposed, we now directly use the memory buffer. I separated this into another commit in case it causes any issues or if there are anticipated future uses for the file-based approach, making it easier to roll back or opt not to merge.

Thank you for the amazing project. ;)

Fixes #878

- [X] I have read `docs/CONTRIBUTING.md`
- [X] I have read `docs/ROADMAP.md`

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/computer/display/display.py
+++ b/interpreter/core/computer/display/display.py
@@ -1,14 +1,11 @@
 import base64
-import os
 import pprint
-import tempfile
 import time
 import warnings
 from io import BytesIO
 
 import matplotlib.pyplot as plt
 import requests
-from PIL import Image
 
 from ..utils.recipient_utils import format_to_recipient
 
@@ -68,8 +65,6 @@ class Display:
             )
             return
 
-        temp_file = tempfile.NamedTemporaryFile(suffix=".png", delete=False)
-
         if quadrant == None:
             # Implement active_app_only!
             if active_app_only:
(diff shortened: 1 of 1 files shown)


KillianLucas:
Great fixes @double-thinker. I'm not sure why I went with the `tempfile` approach! This is much smarter, and the conversion sounds like a solid precaution/solution. Merging.

PR #896, opened 2024-01-10
Fixed %tokens command

CyanideByte:
Just updated variables to the new names to fix a crash

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


Notnaton:
Nice catch!
We should add all commands to tests @KillianLucas?

KillianLucas:
Looks great! Many thanks @CyanideByte. I would have never caught this, and it totally breaks %tokens without your patch.

And I agree @Notnaton— though I'm not sure how to test %magic commands since they only exist for the terminal interface... I think we'll need to use a subprocess to run `interpreter` and test them (but they'll be hard to verify if they've passed...)

PR #935, opened 2024-01-17
Added the Ability to edit the code before running it.

Eliaskhal:
Now the user can edit the code before running it. 
OI does not keep track of previous code (before it was edited).

#### What I did exactly:
- Instead of just asking the user if they want to run the code or not, I added a while loop that keeps on asking if they want to run the code or edit it. The purpose of using the loop is to give the user the ability to re-edit the code until they are satisfied.
- When the user chooses to edit the code, the code gets saved in a temp file that is opened in "VS Code". After the user is done editing, they can type any character into the terminal and the "code" variable will be changed.

#### Possible Improvements:
- The code before it's edited is not saved in the message history.
- It would be better if all previous versions of the code were saved so the user can go back to each one if they want to.

#### NOTE: 
I haven't been able to test the code since I had problems running it even though I have followed the documentation.

#### Check List:

Tests:
- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/terminal_interface/terminal_interface.py
+++ b/interpreter/terminal_interface/terminal_interface.py
@@ -10,6 +10,7 @@ except ImportError:
 
 import os
 import platform
+import tempfile
 import random
 import re
 import subprocess
@@ -41,6 +42,22 @@ try:
 except:
     # If they don't have readline, that's fine
     pass
+
+def edit_code(code):
+    with tempfile.NamedTemporaryFile(mode='w', delete=False, suffix=".py") as temp_file:
+        temp_file.write(code)
+        temp_filename = temp_file.name
+
+    subprocess.run(['code', temp_filename], check=True)
+
+    input("Press Enter when you have finished editing...")
+
+    with open(temp_filename, 'r') as modified_file:
+        modified_code = modified_file.read()
+
+    os.remove(temp_filename)
+
+    return modified_code
 
 
 def terminal_interface(interpreter, message):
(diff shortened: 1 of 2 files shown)


Eliaskhal:
I also think it would be a good idea to give the user the ability to choose their favorite editor by asking them once the first time they want to edit a code.

interpreter/terminal_interface/terminal_interface.py
```diff
@@ -42,6 +43,22 @@
     # If they don't have readline, that's fine
     pass
 
+def edit_code(code):
+    with tempfile.NamedTemporaryFile(mode='w', delete=False, suffix=".py") as temp_file:
+        temp_file.write(code)
+        temp_filename = temp_file.name
+
+    subprocess.run(['code', temp_filename], check=True)
```
Notnaton, line 51: Use something like this to open it in default editor. Check language and add extention to the tmp file. python = .py etc..
```
import os
import subprocess
import platform

def open_file_default_editor(file_path):
    try:
        if platform.system() == 'Windows':
            os.startfile(file_path)
        elif platform.system() == 'Darwin':  # macOS
            subprocess.call(('open', file_path))
        elif platform.system() == 'Linux':
            subprocess.call(('xdg-open', file_path))
        else:
            print(f"Unsupported OS: {platform.system()}")
    except Exception as e:
        print(f"Error opening file: {e}")
```

Notnaton (changes requested):

Notnaton:
https://github.com/KillianLucas/open-interpreter/pull/830
Here is a similar PR, but outdated

interpreter/terminal_interface/terminal_interface.py
Eliaskhal: Thank you for your feedback!
For checking the language, does OI already know what language the code he wrote is in?

Notnaton: Yes, but if language is None it defaults to python

Eliaskhal: Can you please give me the name of the variable that holds that value?

One more thing, I noticed that  whatever changes I made to the code, when I run interpreter in the terminal, it just runs the first version. Why is that?

Notnaton: Have a look at contributors.md

KillianLucas:
Hi @Eliaskhal!

This is going to be such a useful feature. I think this is close, but I want to approach it in a slightly different place.

My main edit: I think it's too much to ask the user if they want to edit _every_ code block, especially to insert it into a familiar (y/n) ask. I think `edit` is much more similar to the other %magic commands (`%shell` is very similar to `edit`, in that it lets you type and run code) than to the central OI concept of running code (which putting it below every code block, as this does, would imply).

I'm happy to merge if it can be converted to a magic command with [the flow outlined here](https://github.com/KillianLucas/open-interpreter/pull/830#issuecomment-1854989795), otherwise I'll merge once this feature becomes my priority, then I'll convert it into a magic command.

Also it's worth noting that we do store the file extension for the code, so if you `interpreter.computer.terminal.get(language).file_extension` you can retrieve the proper file extension, allowing you to open it in the user's default editor.

Great work on building the feature. I'm personally very excited to use this, and your PR works incredibly well.

Eliaskhal:
Hey @KillianLucas, I did not notice your comment and I ended up adding what @Notnaton suggested. I would be more than happy to use your idea. My only problem is that I am still having issues understanding how everything is architected in the project. 

I understand OI has a message history and magic commands. If you don't mind, could you please explain a bit about how those messages are stored? and where does the user type the magic command exactly?

Eliaskhal:
I think I was able to figure it out. I added the magic command and it works now. I will close this request and start a new one. Thank you!

Issue #875, opened 2024-01-06
NameError: name 'computer' is not defined  


youssefabdelm:
### Describe the bug

When I run `interpreter --os`

And then attempt a command like:
`Play a boiler room set on youtube`

I get a `NameError`:

```
▌ OS Control enabled                                                                                                                                                                

To find items on the screen, Open Interpreter has been instructed to send screenshots to api.openinterpreter.com (we do not store them). Add --offline to attempt this locally.       

Make sure that screen recording permissions are enabled for your Terminal or Python environment.                                                                                      

Warning: In this mode, Open Interpreter will not require approval before performing actions. Be ready to close your terminal.                                                         

> Play a boiler room set on youtube
                                                                                                                                                                                      
  The first step is to open YouTube. We can do this by using a keyboard hotkey to open Spotlight and then typing "Safari" to open the browser. Once Safari is open, we can navigate   
  to YouTube by writing the URL in the address bar. Let's start by opening Safari.                                                                                                    
                                                                                                                                                                                      
Expecting value: line 1 column 1 (char 0)
Continuing.

                                                                                                                                                                                      
  computer.keyboard.hotkey("space", "command") # Opens spotlight                                                                                                                      
  computer.keyboard.write("Safari")                                                                                                                                                   
  computer.keyboard.press("enter") # To open Safari                                                                                                                                   
                                                                                                                                                                                      
                                                                                                                                                                                      
  ---------------------------------------------------------------------------                                                                                                         
  NameError                                 Traceback (most recent call last)                                                                                                         
  Input In [4], in <cell line: 2>()                                                                                                                                                   
        1 print('##active_line1##')                                                                                                                                                   
  ----> 2 computer.keyboard.hotkey('space', 'command')                                                                                                                                
        3 print('##active_line2##')                                                                                                                                                   
        4 computer.keyboard.write('Safari')                                                                                                                                           
                                                                                                                                                                                      
  NameError: name 'computer' is not defined                                                                                                                                           
                                                                                                                                                                                      
                                                                                                                                                                                      
  It seems there was a mistake in my attempt to run the code. "computer" is not recognized as defined within the current execution environment. Let's correct this by using the       
  appropriate syntax to execute the code on your machine. We will try to open Safari using the correct approach. Let's run the correct code.                                          
                                                                                                                                                                                      
Expecting value: line 1 column 1 (char 0)
Continuing.

                                                                                                                                                                                      
[...]

PR #937, opened 2024-01-17
Fix: NameError: name 'computer' is not defined

Arrendy:
The `run` method in the `Terminal` class is a generator function that executes some code and yields the results one at a time. In a previous commit, the for loop was removed that was used to iterate over these results, which caused the `run` method to not actually be executed because its results were not being consumed. This commit adds the for loop back to ensure that the `run` method is properly executed.

https://github.com/KillianLucas/open-interpreter/issues/875

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/terminal_interface/start_terminal_interface.py
+++ b/interpreter/terminal_interface/start_terminal_interface.py
@@ -1,7 +1,6 @@
 import argparse
 import os
 import platform
-import re
 import subprocess
 import sys
 import time
--- a/interpreter/terminal_interface/start_terminal_interface.py
+++ b/interpreter/terminal_interface/start_terminal_interface.py
@@ -541,11 +541,12 @@ Include `computer.display.view()` after a 2 second delay at the end of _every_ c
         #         print(chunk.get("content"))
 
         # Give it access to the computer via Python
-        interpreter.computer.run(
+        for _ in interpreter.computer.run(
             "python",
             "import time\nfrom interpreter import interpreter\ncomputer = interpreter.computer",  # We ask it to use time, so
             display=args.verbose,
-        )
+        ):
+            pass
 
         if not args.auto_run:
             display_markdown_message(
--- a/interpreter/terminal_interface/start_terminal_interface.py
+++ b/interpreter/terminal_interface/start_terminal_interface.py
@@ -299,7 +299,7 @@ def start_terminal_interface(interpreter):
 
         config_directory = os.path.dirname(config_file)
 
-        print(f"Opening config directory...")
+        print("Opening config directory...")
 
         if platform.system() == "Windows":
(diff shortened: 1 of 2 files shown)


Notnaton:
  File "<string>", line 1, in <module>
  File "E:\dev\open-interpreter\interpreter\core\core.py", line 46, in start_terminal_interface
    start_terminal_interface(self)
  File "E:\dev\open-interpreter\interpreter\terminal_interface\start_terminal_interface.py", line 543, in start_terminal_interface
    for _ in interpreter.computer.run(
  File "E:\dev\open-interpreter\interpreter\core\computer\terminal\terminal.py", line 40, in run
    for chunk in self._streaming_chat(language, code, stream=True):
                 ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Terminal' object has no attribute '_streaming_chat'

Notnaton:
Added a quick fix, just putting `stream=True`

interpreter/terminal_interface/start_terminal_interface.py
```diff
@@ -541,11 +540,13 @@ def start_terminal_interface(interpreter):
         #         print(chunk.get("content"))
 
         # Give it access to the computer via Python
-        interpreter.computer.run(
-            "python",
-            "import time\nfrom interpreter import interpreter\ncomputer = interpreter.computer",  # We ask it to use time, so
+        for _ in interpreter.computer.run(
```
KillianLucas, line 543: This is actually why I added that part that pulls from the stream! It allows you to just run `interpreter.computer.run` without the `for _ in` syntax. I'll merge but will revert this.

interpreter/terminal_interface/start_terminal_interface.py
```diff
@@ -630,7 +631,7 @@ def start_terminal_interface(interpreter):
 
     # Set attributes on interpreter
     for argument_name, argument_value in vars(args).items():
-        if argument_value != None:
+        if argument_value is not None:
```
KillianLucas, line 634: ?

KillianLucas:
Thanks @Arrendy! And thanks @Notnaton for working on this as well! This was totally a mistake I made, luckily committed after the last version that went out to `pip`, so just a problem for us working off the repo.

Renaming it to `run` is exactly how it ought to work— if `stream=False`, we just pull from an identical `run` with `stream=True` and return that output. Merging!

interpreter/terminal_interface/start_terminal_interface.py
Arrendy: https://peps.python.org/pep-0008/#programming-recommendations

"Comparisons to singletons like None should always be done with is or is not, never the equality operators."

PR #947, opened 2024-01-19
create new cli entry point and update poetry script

Arrendy:
- created a new file `cli.py` that is the terminal interface entry point
- removed the logic that served this from `core.py`
- updated poetry script to call the new function

I tested it as a CLI app and as an imported library and both worked as expected without changing the user experience. Since the user experience didn't change, I didn't update any docs but if the inner workings are documented publicly, let me know and I can update them.

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

Note on tests, in order to test that importing still works, I did:
```
import sys
sys.path.insert(0, "/Users/mike/Code/open-interpreter")

from interpreter import interpreter


def run(prompt, key):
    interpreter.llm.api_key = key
    output = interpreter.chat(prompt, stream=True)
    for chunk in output:
        print(chunk)


run("hello", "sk-XXXX")
```


KillianLucas:
Oh, nice! Appreciate you reading that comment and fixing this Mike.

Is there any way we can put it into the /terminal_interface instead of /interpreter? Can poetry handle that..? Let me know if not, I'll still merge.

I really struggle with relative import stuff so thank you for separating it out— this already makes much more sense to me.

Arrendy:
> Oh, nice! Appreciate you reading that comment and fixing this Mike.
> 
> Is there any way we can put it into the /terminal_interface instead of /interpreter? Can poetry handle that..? Let me know if not, I'll still merge.
> 
> I really struggle with relative import stuff so thank you for separating it out— this already makes much more sense to me.

Done! 

Does the file/function naming still make sense to you now that it's within the `terminal_interface` directory?

KillianLucas:
Nice! This makes me realize we should just fold this into start_terminal_interface— that should be our poetry entry-point, and it should create `interpreter` just as this beautiful `cli.py` does. I'll accept the PR then make that change.

Great work, merging now!

PR #982, opened 2024-01-30
Move get_storage_path to further decouple Core from Terminal Interface

MikeBirdTech:
One more step towards decoupling `Core` from `Terminal Interface` by moving `get_storage_path` util into `Core`.

`OpenInterpreter` objects are initialized with a `conversation_history_path` field which is assigned the return of `get_storage_path('conversations')` so it makes sense to have this util within `Core` instead of `Terminal Interface`.

`interpreter/terminal_interface/conversation_navigator.py` can use `interpreter.conversation_history_path` instead of calling `get_storage_path('conversations')` since the `interpreter` instance will have already determined this value.

`interpreter/terminal_interface/utils/get_conversations.py` wasn't being used - deleted it.

However, `interpreter/terminal_interface/utils/profiles.py` is now coupled to `core` because of this change. I think a bigger conversation around profiles is needed. Is it enough that `Terminal Interface` has the `oi_dir` util? If needed, can other directories be generated here? If `Profiles` are configuring `OpenInterpreter` fields, should they be moved to `Core` too? 

`interpreter/terminal_interface/utils/ARCHIVE_get_config.py` is now coupled to `core` but with this util being archived, I didn't think it was worth a deeper investigation.

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


ianjennings:
@testdriverai open a new terminal window and run the command `interpreter`. After 5 seconds, write the text "open google chrome" in the terminal window and hit return. If google chrome opens, the task is successful.

testdriverai:
I ran your tests @ianjennings and here's what I found:
["TestDriver.ai](https://app.dashcam.io/replay/65ba84fdd1d1d60063d52d02?share=laSsm11lqaW5CeiAd35wtg)

Notnaton:
Tested on windows,
updated it to upstream/main

Notnaton (approved):

PR #996, opened 2024-02-04
added string to multiple python file generation for skill library

akuten1298:
Added functionality to convert from string into python code and store in a dict

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


KillianLucas:
Incredible work Aakash! Will merge if you want to move this function into `interpreter/core/computer/terminal/languages/jupyter_language.py` — I put a function at the at there called `string_to_python`, if you populate it with your code we'll be one step closer.

I think it should be changed to just handle the conversion instead of the saving tho— basically ~ line 102, I think it should just return a dictionary in this format:

`{"function_name": "import whatever\ndef function_name(): ..."}`

Where the key is the function name and the value is the code. Then OI will handle saving it to the right place. Let me know if that makes sense!

akuten1298:
Got it, sure. I've added that change and updated the PR, lmk if this is exactly what you are looking for. Attaching an image here to make it easier to check.

Screenshot 2024-02-04 at 7 24 58 AM

KillianLucas:
Forgot to thank you for this after merging Aakash, brilliant work. Also we're still at the Airbnb if you're free tomorrow :)

PR #1011, opened 2024-02-13
Add doc strings to all computer API functions for the LLM to use in the skill library

tyfiero:

Added doc strings to all computer API functions for use by the skill library. These short doc strings will be used to semantically search for the most relevant computer API function to use.

Note: I removed "Args" inside of the existing doc strings, it sounds like we will programmatically inject that information into the skill library instead of writing it all out, as far as I understand. Happy to add them back or add Args to all doc strings if necessary :)

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/computer/browser/browser.py
+++ b/interpreter/core/computer/browser/browser.py
@@ -6,6 +6,9 @@ class Browser:
         self.computer = computer
 
     def search(self, query):
+        """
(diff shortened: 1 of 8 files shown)


KillianLucas:
Awesome, thanks Ty! Super helpful to the 01 + OI projects and to many language models in the future lol. Merging!

Issue #990, opened 2024-02-02
Make the application launch faster.


Steve235lab:
### Is your feature request related to a problem? Please describe.

The application launch a bit too slow especially for cases that users just want some basic functions like `-h`, `--conversations` and so on.

### Describe the solution you'd like

I find that the entry point for current version is:
```python
def main():
    interpreter = OpenInterpreter()
    try:
        start_terminal_interface(interpreter)
    except KeyboardInterrupt:
        pass
```
which means the `OpenInterpreter` instance was initialized first before processing the arguments.

I think we can have two enumerates for those arguments need the `OpenInterpreter` instance to work and those don't, so we can avoid initializing the instance and have a faster application.

### Additional context

I will open a pull request to do so in the near future.

Steve235lab:
I have tried to init some heavy objects with multithreads, here's the code: [my fork, just check the latest commits](https://github.com/Steve235lab/open-interpreter/tree/async-init-interpreter-instance) However, it doesn't work well. I run a simple benchmark to measure the launch time:
1. Write a benchmark script `time_this.sh`:

   ```bash
   #!/bin/bash
   
   time_this() {
     START=$(python -c "import time; print(int(time.time() * 1000))")
     $@
     END=$(python -c "import time; print(int(time.time() * 1000))")
     DIFF=$(( $END - $START ))
     echo "Execution time: $DIFF milliseconds"
   }
   ```

2. Source the bash script:

   ```bash
   chmod +x time_this.sh && source time_this.sh
   ```

3. Then run benchmarks:

   ```bash
   time_this interpreter -h
   ```

With this simple benchmark we can get the time cost to launch the app, and compare the difference after modifications.

So I think that I'd better locate lines that take the most time to run first. Then I generated a performance report `performance.txt` with `cProfile`:

```bash
python -m cProfile -o performance.pstats interpreter -h
```

Here's the file: [performance.txt](https://github.com/KillianLucas/open-interpreter/files/14152271/performance.txt)
Please be noticed that this is not a real txt file, but a `.pstats` file which is not a supported format on GitHub comments. To read this file, you need `pstats` which is a module from python standard library.

**Are there anyone can help me with this problem? Or I had to finish this alone?**

Steve235lab:
@yuhanj2 HELP!!!

tyfiero:
[...]

PR #1020, opened 2024-02-16
Boost performance by lazy loading modules

tyfiero:

Lazy load the computer API modules to boost start-up speed. In my tests, it boosted startup speed nearly 3X. The computer API modules rely on a lot of C code, which were being imported if you had them installed. The first time you install open-interpreter (or when you use a new venv), it still takes a while, but subsequent startups are faster. 

I am happy to modify any code here. 

This boosting of performance is a never-ending pursuit, I can't say that it "Fixes"  issue #990, but it certainly helps. There is more work to be done to improve performance. When running "--os", the subprocess has to re-import the computer API, which doubles the startup time. I'm sure this can be improved but is not addressed in this PR.

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


KillianLucas:
A LEGEND.

This has PLAGUED the package. Major usability problem, brilliant fix, instant merge. 

Incredible work Ty!

Issue #1046, opened 2024-02-29
Multi-line Input Support


Steve235lab:
### Is your feature request related to a problem? Please describe.

OI doesn't support multi-line input from the CLI. It's difficult for user to input code pieces directly from the terminal.

### Describe the solution you'd like

Find a good way to handle multi-line input.

Here are some potential solutions:
* Add some logic when processing user's input, if get line ending like `\n`, search for some specific combo in the line like ` ``` ` , if found then await for more input until get another ` ``` `.
* User write the code into a file and let the OI read the file (this one has been already done).

Steve235lab:
Any good idea? I might try the first solution in the next few days.

Notnaton:
There is a pr draft of future gui using textual, that should handle multi line input.

Steve235lab:
> There is a pr draft of future gui using textual, that should handle multi line input.

Yes, I saw the PR before, it's great and I'll definitely try the GUI app in the future. However it's not exactly the thing I want right here. In my opinion OI is special and powerful largely because it's the only usable CLI Chatbot client out of box, so enhance the CLI user experience is quite important for me. I'll find a good solution later.

PR #1048, opened 2024-03-01
Multi-line input support.

Steve235lab:
Support multi-line input by defining the function `cli_input`, which detects special combo (`start_marker`) in user's input like ` ``` ` to await more input until get an another special combo (`end_marker`). In this way, users can input multiple lines at one single input, which is very helpful when the users try to chat about pieces of code copied from their text editors. To enable this new feature, you can use `interpreter --multi_line` or `interpreter -ml` to launch OI, or you can simply set `multi_line: True` in your profile.

Although this new feature is ready to go for most cases, there's still some improvements can be done:
1. If the multi-line part contains the `end_marker` string, it will cause an unexpected ending of the current input, which means you can't have ` ``` ` contained in your multi-line inputs.
2. This new feature can be mis-triggered and get users trapped in the multi-line input loop. (This one is not a problem for now because this new feature is disabled by default in args.)

Here's an example, the first conversation was launched with this new feature disabled by default, and the second one was launched with this new feature enabled:
截屏2024-03-02 01 26 50

#1046 

- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/docs/usage/terminal/arguments.mdx
+++ b/docs/usage/terminal/arguments.mdx
@@ -16,7 +16,7 @@ title: Arguments
 
 **[Options](/docs/usage/terminal/arguments#options)**
 
-`--safe_mode`, `--auto_run`, `--force_task_completion`, `--verbose`, `--max_budget`, `--speak_messages`.
+`--safe_mode`, `--auto_run`, `--force_task_completion`, `--verbose`, `--max_budget`, `--speak_messages`, `--multi_line`.
 
 **[Other](/docs/usage/terminal/arguments#other)**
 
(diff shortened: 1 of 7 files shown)


Steve235lab:
> I wonder if we can detect when the user is pasting something in, and just enable this immediately until the paste is complete somehow? Maybe we could turn on input() for like 100ms after a newline happens, so that if someone pastes something in, it's under the cutoff and the whole thing gets sent in. But if they hit `enter` normally, they just experience a 100ms delay.

Great idea, I'll implement this later.

pinggit:
what is the solution now? I still can't ask any real questions about my code which is of course multiple lines

cschubiner:
@KillianLucas  this no longer works, yeah? 
image

PR #1093, opened 2024-03-18
does not prompt openapi key when running --api_base

### LINKED ISSUE #1076: When using LM studio, user is prompted to enter their OpenAI API key when this value isn't needed
### Describe the bug

The change in `--local` mode now requires users to launch OI with `interpreter --api_base http://localhost:1234/v1` if they want to use LM Studio. When they run this command, they are prompted to enter their OpenAI API key. Pressing enter without typing any value allows OI to work with LM Studio successfully. However, the user should not be prompted to enter their OpenAI API key when they are using a local model.

### Reproduce

1. Disable `OPENAI_API_KEY` environment variable
2. Run `interpreter --api_base http://localhost:1234/v1`
3. Observe OI asking for OpenAI API key

### Expected behavior

Running `interpreter --api_base http://localhost:1234/v1` should work without prompting user for their OpenAI API key.

### Screenshots

Screenshot 2024-03-13 at 11 41 50 AM

### Open Interpreter version

0.2.2

### Python version

3.11.3

MacOS 14.3

### Additional context

_No response_

lorenzejay:

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/terminal_interface/validate_llm_settings.py
+++ b/interpreter/terminal_interface/validate_llm_settings.py
@@ -29,7 +29,7 @@ def validate_llm_settings(interpreter):
 
             # OpenAI
             if interpreter.llm.model in litellm.open_ai_chat_completion_models:
-                if not os.environ.get("OPENAI_API_KEY") and not interpreter.llm.api_key:
+                if not os.environ.get("OPENAI_API_KEY") and not interpreter.llm.api_key and not interpreter.llm.api_base:
                     display_welcome_message_once()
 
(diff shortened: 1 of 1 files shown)


KillianLucas:
Great fix, very clean! 

Long term thoughts: Tbh that whole file seems wrong to me. We should probably be letting people do whatever, then we catch certain errors and display them in a more friendly, proactive way (e.g. missing OpenAI key? We prompt the user for it).

Issue #915, opened 2024-01-14
--context_window & --max_tokens issue


chuangzhidan:
### Describe the bug

**run into this problem after "interpreter --model openai/gpt-4-1106-preview --api_key sk-j5axpBIXNNHI3WTSXXXXXXXXXXX016B046"**
▌ Model set to openai/gpt-4-1106-preview
Open Interpreter will require approval before running code.
Use interpreter -y to bypass this.
Press CTRL-C to exit.
**> hi**
We were unable to determine the context window of this model. Defaulting to 3000.
If your model can handle more, run interpreter --context_window {token limit} --max_tokens {max tokens per response}.
Continuing...
·····

**so i set the value as suggested ,and i got this .do not know the right format and value ,if you know  what i mean**

usage: interpreter [-h] [-ci CUSTOM_INSTRUCTIONS] [-s SYSTEM_MESSAGE] [-y] [-v] [-m MODEL] [-t TEMPERATURE] [-lsv] [-lsf] [-c CONTEXT_WINDOW] [-x MAX_TOKENS] [-b MAX_BUDGET] [-ab API_BASE] [-ak API_KEY] [-av API_VERSION]
                   [-xo MAX_OUTPUT] [-fc] [-dt] [-o] [-sm] [-safe {off,ask,auto}] [-cf CONFIG_FILE] [-f] [-l] [-vi] [-os] [--config] [--reset_config] [--conversations] [--version]
interpreter: error: argument -c/--context_window: invalid int value: '{10000}'

**by the way  , i do not need to input api-key every time i open up a terminal ,yes? tried the stex too ,did not do do the trick ,still have to set api-key every time**

much appreciated if you can help me with this!

### Reproduce

interpreter --model openai/gpt-4-1106-preview --api_key sk-j5axpBIXNNHI3WTSXXXXXXXXXXX016B046
interpreter --model openai/gpt-4-1106-preview --context_window {10000} --max_tokens {5000} --api_key   sk-j5axpBIXNNHI3WTSXXXXXXXXXXX016B046
interpreter --m/gpt-4-1106-preview --c {10000} --x {5000} --ak sk-j5axpBIXNNHI3WTSXXXXXXXXXXX016B046

### Expected behavior

expect no reporting error  like max token and context widow issue

### Screenshots

![Uploading $2SJXQJZ9MDVVD6MI$CF{(K.png…]()

### Open Interpreter version

0.2.0

### Python version

3.9

### Operating System name and version

windows11

[...]

PR #1069, opened 2024-03-12
Recognize gpt-4-0125-preview

mmazur:
### Describe the changes you have made:

There's more than one gpt 4 turbo preview models, both should be recognized as valid.

### Pre-Submission Checklist (optional but appreciated):

- [ ] I have included relevant documentation updates (stored in /docs)
- [ ] I have read `docs/CONTRIBUTING.md`
- [ ] I have read `docs/ROADMAP.md`

### OS Tests (optional but appreciated):

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux

PR #1099, opened 2024-03-19
Support OpenAI's new models.

Steve235lab:
Add default context window and max tokens configs for OpenAI's new models: `gpt-4-turbo-preview`, `gpt-4-0125-preview`, and `gpt-4-1106-vision-preview`.

If we can keep these configs updated with as more models as possible, maybe can avoid issues like #915

- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/interpreter/terminal_interface/start_terminal_interface.py
+++ b/interpreter/terminal_interface/start_terminal_interface.py
@@ -353,7 +353,16 @@ def start_terminal_interface(interpreter):
 
     ### Set some helpful settings we know are likely to be true
 
-    if interpreter.llm.model == "gpt-4-1106-preview":
+    if interpreter.llm.model in [
+        "gpt-4-turbo-preview",
+        "gpt-4-0125-preview",
+        "gpt-4-1106-preview",
+        "gpt-4-1106-vision-preview",
+        "openai/gpt-4-turbo-preview",
+        "openai/gpt-4-0125-preview",
+        "openai/gpt-4-1106-preview",
+        "openai/gpt-4-1106-vision-preview",
+    ]:
         if interpreter.llm.context_window is None:
             interpreter.llm.context_window = 128000
         if interpreter.llm.max_tokens is None:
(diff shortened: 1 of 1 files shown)


Steve235lab:
Here's the reference [Models - OpenAI API](https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo).

Steve235lab:
I don't use other LLMs much, so my commit only focus on OpenAI GPTs. If anyone familiar with other models and their proper context window and max tokens configs, please add configs like this commit to make OI support more models by default.

Steve235lab:
My bad, haven't seen there's already one https://github.com/KillianLucas/open-interpreter/pull/1069

Notnaton:
There is probably a better way to do this. I'll take a look later today.

CyanideByte:
Just a note, for the vision models, function calling has to be False or they break.

Steve235lab:
> There is probably a better way to do this. I'll take a look later today.

Maybe we can simply use `str.startwith` to match the prefix of models like `gpt-4` and `openai/gpt-4`?

Steve235lab:
Strongly recommend to try `gpt-4-turbo-preview` model, it's cheaper and faster.

CyanideByte:
> Strongly recommend to try `gpt-4-turbo-preview` model, it's cheaper and faster.

Exactly this. In OpenAI's January blog post, they stated that 70% of their customers had transitioned to gpt-4-turbo from gpt-4.
Its probably higher by now, gpt-4 is a 9 month old legacy model at this point.

Steve235lab:
@KillianLucas If you are going to merge this, just squeeze my commits. Nobody wants 3 lines changed in 4 commits. >_<

KillianLucas:
Nice @Steve235lab, makes the most of these models without needing to set specific settings.

Will do another test run soon (and set up some more robust testing) — last time I tested with the demos from the launch video, `gpt-4` was unfortunately much more intelligent/robust than `gpt-4-turbo`. If that's changed, we should immediately switch.

Merged! Unfortunately I'm not very good at git, not sure how to squeeze commits, but let me know if it's important to do and I'll figure it out! Thank you so much for this PR.

PR #1105, opened 2024-03-21
Feature/ruby support

bars0um:
Added Ruby Language support

https://github.com/OpenInterpreter/open-interpreter/issues/1094

- [ n/a? ] I have included relevant documentation updates (stored in /docs)
- [ x ] I have read `docs/CONTRIBUTING.md`
- [ x] I have read `docs/ROADMAP.md`

- [ x ] Tested on Linux

### DIFF
--- a/interpreter/core/computer/terminal/languages/ruby.py
+++ b/interpreter/core/computer/terminal/languages/ruby.py
@@ -1,0 +1,144 @@
+
+
+import re
+from pathlib import Path
+from .subprocess_language import SubprocessLanguage
+
+
+class Ruby(SubprocessLanguage):
+    file_extension = "rb"
+    name = "Ruby"
+
+    def __init__(self):
+        super().__init__()
+        self.start_cmd = ["irb"] 
+
+    def preprocess_code(self, code):
+        """
+        Add active line markers
+        Wrap in a tryCatch for better error handling 
+        Add end of execution marker
+        """
+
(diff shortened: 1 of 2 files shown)


MikeBirdTech:
Thanks for the pr @bars0um!

I'm not a ruby guy, so I was wondering if you could include a quick screencast showing how the changes work with Ruby?

KillianLucas:
A screencast would be amazing! But also this doesn't appear to be a breaking change, just additive. If it doesn't work on some systems we can fix from there. Merged, great work @bars0um on totally understanding how to add new languages with no docs on that!

bars0um:
Here's a brief clip of this change in action.

https://github.com/OpenInterpreter/open-interpreter/assets/39853587/2981b4d6-fb23-412b-8c3a-79c4bde42e63

This allows OI to create ruby code and execute it. Not sure what the use-case would be, in my case I'm trying to make it write ruby code into a file. So in a fork i have, I basically treat Ruby as a BaseLanguage instead and write the output to a file named according to a comment at the top of the file that I direct the language model to put at the top via the system instructions. The approach appears to be promising. I'm looking into how to get the code re-read and for the LLM to iterate a few times on it to improve...that's proving a bit challenging as it kinda goes off into an endless discovery of what life, the universe and everything...falling miserably short of identifying the ultimate answer...that of course being 42

PR #1121, opened 2024-03-24
[FIX] Broken Link to Setup Fixed

Sandeepsuresh1998:

Don't need relative path if in same dir, so updated to just filename :)

- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux


KillianLucas:
Nice! I think because we run our docs on mintlify.com we won't need the file extension (it serves an HTML page instead of the base MDX) but we should be able to get rid of the /getting-started. Merged with that change.

PR #1132, opened 2024-03-26
Check for GPU or MPS availability before using CPU

jcp:

In `point.py`, check for GPU or MPS before defaulting to CPU.

- Fixes https://github.com/OpenInterpreter/open-interpreter/issues/1131

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/computer/display/point/point.py
+++ b/interpreter/core/computer/display/point/point.py
@@ -490,7 +490,13 @@ if fast_model == False:
     # embeddings = embed_images(images, model, transforms)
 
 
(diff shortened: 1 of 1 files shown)


KillianLucas:
Thanks for this @jcp! Let's test it before pushing, I think the reason we forced it to CPU was because the hashes that point to the vectors get stored on one of those devices, and we kept hitting an error where they weren't on the same device as the query vector. If that happens again I might revert this, but a part of me thinks that problem came from elsewhere. Merging!

Issue #1151, opened 2024-03-28
`anonymized_telemetry: false` not working in profile and env


LucienShui:
### Describe the bug

When I'm trying to disable telemetery using the profile, I find that interpreter still calling the posthog.com.

### Reproduce

Here's my `profiles/default.yaml`.

```yaml
llm:
  model: "model"
  temperature: 0
  api_key: "***"
  api_base: "***"
  context_window: 2048
  llm_supports_functions: false
  llm.llm_supports_vision: false
anonymized_telemetry: false

version: 0.2.1
```

After that when I execute `interperter` in the terminal, I got error of request `posthog.com`. (Cause my AdGuardHome has banned it)

### Expected behavior

No telemetry.

### Screenshots

<img width="732" alt="image" src="https://github.com/OpenInterpreter/open-interpreter/assets/30151093/18edfde6-1283-4eaa-9aba-a1f6554baa0b">

### Open Interpreter version

0.2.4

### Python version

3.10.14

### Operating System name and version

macOS 14.3.1

### Additional context

`interpreter --disable_telemetry` worked.

bloomsa:
I'm not sure about the history of the telemetry config in `profiles/default.yaml` but as of the current docs, it looks like the proper key is `disable_telemetry`, i.e. `disable_telemetry: true`
https://docs.openinterpreter.com/settings/all-settings#disable-telemetry

LucienShui:
> I'm not sure about the history of the telemetry config in `profiles/default.yaml` but as of the current docs, it looks like the proper key is `disable_telemetry`, i.e. `disable_telemetry: true` https://docs.openinterpreter.com/settings/all-settings#disable-telemetry

@bloomsa Maybe you're right but still not working for me.

<img width="1056" alt="image" src="https://github.com/OpenInterpreter/open-interpreter/assets/30151093/caaa3e08-755d-4161-ad5c-9de53728bf20">

BTW, there is another place about how to disable telemetry, and it can be found in the current `main` branch as shown bellow:

<img width="1344" alt="image" src="https://github.com/OpenInterpreter/open-interpreter/assets/30151093/b55e574d-8c87-495d-b879-67927d7600b1">

https://github.com/OpenInterpreter/open-interpreter/blob/cc6291f8372c9c61cb53f7c1d4e6ef819b8457eb/docs/telemetry/telemetry.mdx?plain=1#L30-L46

LucienShui:
[...]

PR #1159, opened 2024-04-01
Fix/profile disable_telemetry not working

LucienShui:
Currently neither `profile` nor `env` can disable telemetry when using command line without `--disable_telemetry`, cause argument parser's `--disable_telemetry` defaults to `True`, which would always overwrite the `interpreter.anonymous_telemetry` to `True`.

https://github.com/OpenInterpreter/open-interpreter/blob/cc6291f8372c9c61cb53f7c1d4e6ef819b8457eb/interpreter/terminal_interface/start_terminal_interface.py#L142-L150

1. Enabling the option to disable telemetry via profile settings or environmental variables.
2. **[Breaking]** The anonymous_telemetry and anonymized_telemetry options have been removed from both the documentation and the codebase.
3. Ensuring that the `--disable_telemetry` flag functions as its name suggests. (Currently when `--disable_telemetry` is true, then the `interpreter.anonymous_telemetry` is also set to true, and vice versa)

As described in #1151, any of `disable_telemetry`, `anonymous_telemetry`, `anonymized_telemetry` not working as expected.

- [ ] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux


KillianLucas:
Great work @LucienShui! I thought the messaging of "disable" in the param name was very confusing, this is a better solution in every way. Will resolve conflicts and merge.

PR #1171, opened 2024-04-03
added type hints

jayashankarjayan:
Code formatting applied using black and isort as per guidelines.

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -108,14 +108,14 @@ class OpenInterpreter:
         self.speak_messages = speak_messages
 
         # LLM
-        self.llm = Llm(self) if llm is None else llm
+        self.llm: Llm = Llm(self) if llm is None else llm
 
         # These are LLM related
         self.system_message = system_message
         self.custom_instructions = custom_instructions
 
         # Computer
-        self.computer = Computer(self) if computer is None else computer
+        self.computer: Computer = Computer(self) if computer is None else computer
         self.sync_computer = sync_computer
         self.computer.import_computer_api = import_computer_api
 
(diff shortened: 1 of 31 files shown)


imapersonman:
These will be really nice to have!  It looks like there are some inconsistencies in the type hinting, though.  For example, `messages` has type `str | None` [as an OpenInterpreter parameter in core](https://github.com/OpenInterpreter/open-interpreter/blob/c28fa47d70a18836953de7a234bb50a92890775f/interpreter/core/core.py#L45), but is being stored as a `list[dict[str, Any]]` [in the OpenInterpreter constructor's first line](https://github.com/OpenInterpreter/open-interpreter/blob/c28fa47d70a18836953de7a234bb50a92890775f/interpreter/core/core.py#L80).  You can turn on type checking using VSCode + Pylance's `Python › Analysis: Type Checking Mode` option, or use [pyright on the command line](https://microsoft.github.io/pyright/#/command-line).

I think it would be really cool to have this merged once the typing is consistent.

jayashankarjayan:
Yeah, the keyword argument `message`

> These will be really nice to have! It looks like there are some inconsistencies in the type hinting, though. For example, `messages` has type `str | None` [as an OpenInterpreter parameter in core](https://github.com/OpenInterpreter/open-interpreter/blob/c28fa47d70a18836953de7a234bb50a92890775f/interpreter/core/core.py#L45), but is being stored as a `list[dict[str, Any]]` [in the OpenInterpreter constructor's first line](https://github.com/OpenInterpreter/open-interpreter/blob/c28fa47d70a18836953de7a234bb50a92890775f/interpreter/core/core.py#L80). You can turn on type checking using VSCode + Pylance's `Python › Analysis: Type Checking Mode` option, or use [pyright on the command line](https://microsoft.github.io/pyright/#/command-line).
> 
> I think it would be really cool to have this merged once the typing is consistent.

Thanks @imapersonman!. The reason why I have kept `messages` type as `str | None` instead of just `str` is that the constructor of `OpenInterpreter` class actually has `messages` as an optional keyword argument, i.e. it can be `None`

PR #1181, opened 2024-04-07
Update links in docs/

Steve235lab:
Correct links pointing to `/docs/settings/all-settings.mdx`, now it can jump to the correct file on GitHub web page.

### DIFF
--- a/docs/settings/profiles.mdx
+++ b/docs/settings/profiles.mdx
@@ -2,7 +2,7 @@
 title: Profiles
 ---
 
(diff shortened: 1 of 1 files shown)


Steve235lab:
Maybe we should check all the links in the documentations and correct the outdated ones. I corrected this one only because I crashed into this, but I worry if there were many like this. Merge it now or wait for some volunteers (I would like to do, but don't have much time now) fixing all links.

Steve235lab:
I will try to write a script to examine whether a link is good (pointing to something really exist) or outdated later.

KillianLucas:
As bizarre as this is, it's unfortunately a quirk of mintlify.com — the GitHub docs won't point to the right place, but docs.openinterpreter.com should (since "/" on the site is "/docs/" on GitHub).

If there's any way we can make both work (relative links..?) that would be ideal, in the meantime I'll close this PR. Feel free to reopen if there's another way to fix this bug!

PR #1213, opened 2024-04-19
Update local profile so it doen't use function calling

Notnaton:
leaving model = gpt4 will result in function calling. 
Most LM Studio models dont use function calling. making it not work

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux


KillianLucas:
Excellent catch, that shouldn't have worked at all before. I changed it to just turn off function calling, which I think makes more sense, but let me know if it's better to change the model instead for any reason. Merged, great work @Notnaton!

Notnaton:
This works too, but changing the model name would be preferred because I'm making another PR to get context length from the model name.

I'll bring the change in the other PR. 👍

PR #1227, opened 2024-04-23
Added conditional support for Windows, computer not found error

### LINKED ISSUE #1225: 'computer' module not found
### Describe the bug

Any question that needs the interpreter to access the features of the computer module end up giving a 
"NameError: name 'computer' is not defined".
This may be due to the codespace in which the interpreter is running the code.

### Reproduce

1. Install latest open-interpreter
2. Run any code that requires computer module i.e
``` python
from interpreter import interpreter

if __name__ == "__main__":
    interpreter.chat("Email John 'Hi'")
```

### Expected behavior

The computer module should be accessible.

### Screenshots

Screenshot 2024-04-23 111251

### Open Interpreter version

0.2.4

### Python version

3.11.7

Windows 11

### Additional context

_No response_

Abdullah-Gohar:
Changed one of the conditional statements to account for the import of computer module.

Fixes #1225 

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/computer/terminal/terminal.py
+++ b/interpreter/core/computer/terminal/terminal.py
@@ -38,7 +38,8 @@ class Terminal:
 
     def run(self, language, code, stream=False, display=False):
         if language == "python":
-            if self.computer.import_computer_api and not self.computer._has_imported_computer_api and "computer" in code:
+            # If the code contains the word "computer", we will import the computer API if it hasn't been imported yet. (Fix #1225)
+            if (self.computer.import_computer_api or not self.computer._has_imported_computer_api) and "computer" in code:
                 self.computer._has_imported_computer_api = True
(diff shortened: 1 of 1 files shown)


KillianLucas:
Hi @Abdullah-Gohar! Thanks for this PR, I think we might want to solve this issue in another way though. Some folks might want to run a lightweight Open Interpreter in Python without the computer API (I think we should remove reference to it in the system message for this reason, so it won't try to use it unless it has access to it).

We could make it simply print instructions for importing the computer API (initializing interpreter with `interpreter = OpenInterpreter(import_computer_api=True)`) if "computer" is spotted in the code, and `import_computer_api` is `False`. Do you think those solutions would work?

PR #1239, opened 2024-04-27
Add llm_drop_params setting to avoid AnthropicException

nobu007:
litellm.drop_params=True can avoid this error(https://github.com/OpenInterpreter/open-interpreter/issues/1240).
Error in chat: AnthropicException - anthropic does not support parameters
I think drop_params is good altanative for debugging or workaround of future errors.

This option allow litellm.drop_params=True.

Related with https://github.com/OpenInterpreter/open-interpreter/issues/1240.

I found "Use Anthropic function calling" in ROADMAP.md.
Is #1240 just not implemaneted?

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux

### Test(works fine)

```
interpreter --model claude-3-haiku-20240307 -y --llm_drop_params
```

### Note1

This error is probably resolved by this PR.

```
Error in chat: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"messages: roles must alternate between \"user\" and \"assistant\", but found multiple \"user\" roles in a row"}}
```

### Note2

"litellm.modify_params" helps activate [here code](https://github.com/BerriAI/litellm/blob/b7c6f0500b35d6e6afd0d09d6b644365da8445b4/litellm/llms/prompt_templates/factory.py#L848) in litellm.
The first message is empty or non user, dummy message append in anthropic model.
It is not important, but helps for me.

### DIFF
--- a/docs/settings/all-settings.mdx
+++ b/docs/settings/all-settings.mdx
@@ -805,4 +805,45 @@ computer.import_computer_api: True
 ```
 
 </CodeGroup>
-````
+
+### LLM Drop Params
+
+Inform llm to drop unknown params like "functions".
+It supports to easy debug of llm errors.
+
+<CodeGroup>
+
+```bash Terminal
+interpreter --llm_drop_params
+```
+
+```python Python
+interpreter.llm_drop_params = True
+```
+
+```yaml Profile
+llm_drop_params: true
+```
+
+</CodeGroup>
+
+### LLM Modify Params
+
+Inform llm to modify params or messages.
+It supports to easy debug of llm errors.
+
+<CodeGroup>
+
+```bash Terminal
+interpreter --llm_modify_params
+```
+
+```python Python
+interpreter.llm_modify_params = True
+```
+
+```yaml Profile
+llm_modify_params: true
+```
+
+</CodeGroup>
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -79,6 +79,7 @@ class OpenInterpreter:
         import_skills=False,
         multi_line=False,
         contribute_conversation=False,
+        llm_drop_params=False,
     ):
         # State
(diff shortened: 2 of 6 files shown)


KillianLucas:
Hey @nobu007, thanks for testing this with Anthropic and making this fix!

Should we just set `litellm.drop_params = True`? It seems like that could never be bad— if LiteLLM doesn't want to pass a param into the LLM, I feel like we should just let it silently drop it. Any side effects of that that you can see?

nobu007:
@KillianLucas 
In my for 2 weeks try, it seems no side effects forme.
drop_params is used around [here](https://github.com/BerriAI/litellm/blob/99e8f0715e0973e65d32d36907cd36215e4a1c6d/litellm/llms/anthropic.py#L110).

Btw, modify_params is also useful for Anthropic.
modify_params is used around [here ](https://github.com/BerriAI/litellm/blob/61143c8b45e3a3a374084db2b0dc2d0f8090e76c/litellm/llms/prompt_templates/factory.py#L822)and [here](https://github.com/BerriAI/litellm/blob/61143c8b45e3a3a374084db2b0dc2d0f8090e76c/litellm/llms/prompt_templates/factory.py#L1056).

How about adding this also?

nobu007:
@KillianLucas
I added "llm_modify_params" also.
Please check it!

docs/settings/all-settings.mdx
```diff
@@ -805,4 +805,45 @@ computer.import_computer_api: True
 ```
 
 </CodeGroup>
-````
```
nobu007, line 808: Note: This ```` that added in this PR looks no meanings. So I delete it.
https://github.com/OpenInterpreter/open-interpreter/pull/1291

nobu007:
@KillianLucas
I rebased.

PR #1246, opened 2024-04-29
Updates to fast profile

jasonkneen:

Updated this to:

- support gpt-4-turbo which should be super fast enough
- improve prompt to avoid confusion with 'Task is done'
- subtle prompt changes
- version number update to avoid messages

- [x ] I have read `docs/CONTRIBUTING.md`
- [x ] I have read `docs/ROADMAP.md`

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -17,9 +17,14 @@ from .computer.computer import Computer
 from .default_system_message import default_system_message
 from .llm.llm import Llm
 from .respond import respond
-from .server import server
 from .utils.telemetry import send_telemetry
 from .utils.truncate_output import truncate_output
+
+try:
+    from .server import server
+except:
+    # Dependencies for server are not generally required
+    pass
 
 
 class OpenInterpreter:


KillianLucas:
Hi Jason! I think 3.5 is still the best option for speed, the default is 4.5-turbo now so I think `--fast` should be faster. (Tbh I think if the user has a Groq API Key then `--fast` should even use Groq)

I also am not sure I understand "If the user has set you a task and it's done has been completed, say 'Ready to code.'". 

If you have experimented, and you experience better results with this system message (I don't use fast mode very much) then I'll merge. Let me know what you think about switching back to 3.5 and perhaps removing that line! 👍

jasonkneen:
ahh ok slight mistake on that line it should be

```If the user has set you a task and you have completed it, and they done, say 'OK, Ready to code.'```

Essentially I found it a bit confusing when it kept saying "The task is done." when I was testing it with "hello" etc, which of course makes sense if it's been told not to chit-chat so whs trying to make it clearer to OI and the user what was going on. Just thought it was a bit friendlier and re-focus user.

Re 3.5 sure I'll revert it and leave as-is.

shamrockmuffin (approved):

PR #1260, opened 2024-05-10
Auto continue

aj47:

Added a flag `auto_continue` which sends back prompt "Continue" to LLM automatically after conversation has initiated.
This is very useful for 7B models as they tell you what they are going to do without doing it, so usually a lot of manual 'continue' prompting is required. 

- [ Y ] I have included relevant documentation updates (stored in /docs)
- [ Y ] I have read `docs/CONTRIBUTING.md`
- [ Y ] I have read `docs/ROADMAP.md`

- [ Y ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/docs/settings/all-settings.mdx
+++ b/docs/settings/all-settings.mdx
@@ -458,6 +458,26 @@ auto_run: true
 
 </CodeGroup>
 
+### Auto Continue
+
+Automatically tell interpreter to "continue" until CTRL+C is pressed.
+
+<CodeGroup>
+
+```bash Terminal
+interpreter --auto_continue
+```
+
+```python Python
+interpreter.auto_continue = True
+```
+
+```yaml Profile
+auto_continue: true
+```
+
+</CodeGroup>
+
 ### Max Budget
 
 Sets the maximum budget limit for the session in USD.
--- a/docs/usage/python/arguments.mdx
+++ b/docs/usage/python/arguments.mdx
@@ -75,6 +75,20 @@ interpreter.auto_run = False  # Require user confirmation (default)
 
 ---
 
+
+---
+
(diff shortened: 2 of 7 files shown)


KillianLucas:
[pasting from discord, for more visibility if anyone has comments]

this is awesome @aj47! I think we do have a version of this flag in the form of `--force_task_completion`, but i've always hated that abstraction. I like `--auto_continue` more, but there is some extra functionality in the "force task completion" idea— we let the model break out of the loop by emitting a certain message:

```python
# current
interpreter.force_task_completion = True
interpreter.force_task_completion_message = "If the task is done, say 'The task is done.'" # This will be sent to the LLM after every response
interpreter.force_task_completion_breakers = ["The task is done."] # The LLM can say this to break out of the loop
```

so you should be able to recreate the `--auto_continue` behavior in Python with:
```python
# current
interpreter.force_task_completion = True
interpreter.force_task_completion_message = "continue"
interpreter.force_task_completion_breakers = [] # No phrases break it out of the loop
```

this will run OI in a loop, and always hit it with the "continue" message. 

I think we could improve this abstraction, and add TUI flag:

```python
# proposed
interpreter.loop = True
interpreter.loop_message = "continue"
```

```bash
# proposed
interpreter --loop 'continue'
```

do those seem like they would work? also any thoughts/vibes on this @cyanidebyte or @mike.bird? that way we could keep the functionality of force_task_completion, but make it simpler and easily become a form of auto continue

aj47:
Thanks @KillianLucas !
I actually had forgotten about `force_task_completion` when I started this.

What do you hate about it. I think I'll just try it out and focus on bettering that rather than have separate flags

PR #1273, opened 2024-05-17
Add .idea to .gitignore

audriu:

- [ ] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,7 +153,7 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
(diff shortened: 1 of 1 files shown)


KillianLucas:
Hi @audriu! I just want to make sure I understand why? Above that it says this:

```
For a more nuclear option (not recommended) you can uncomment the following to ignore the entire idea
```

I feel if they don't recommend it, we shouldn't do it. What do you think?

audriu:
Well, that comment is just weird it includes the link to a template but does not include anything from it.
Usually, IDE configuration is not committed and I can see it is not committed to open-interpreter either.
Big projects just exclude IDE configurations completely like here: https://github.com/numpy/numpy/blob/main/.gitignore#L9
, https://github.com/pytorch/pytorch/blob/main/.gitignore#L227 , https://github.com/python/cpython/blob/main/.gitignore#L36

PR #1302, opened 2024-06-11
Codestral tts

Notnaton:
Added tts for codestral profile

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux


MikeBirdTech:
Video of profile in action: https://www.youtube.com/watch?v=l9QKGc4AL60

KillianLucas:
NICE!! Love the idea of a local voice assistant. I do think we should keep OI's main repo focused on text, but a profile is a great place to add this extension.

Could the other new files perhaps be pushed into the profile? For example, I think we shouldn't have everyone installing those dependencies, maybe just do a `try except` around the imports of them, then if that fails, you could print a message saying "you need to pip install these dependencies". Then I think the live_transcribe could also go in the profile, at the top. What do you think?

Notnaton:
> Could the other new files perhaps be pushed into the profile? 

Maybe profiles/utils/ is a good spot glue needed for awesome profiles

PR #1332, opened 2024-07-04
Linux support

PiyushDuggal-source:
I made changes to clipboard.py file where we were using `os.name` to detect the OS and setting `modifer_key` which could lead to issues in Linux environment, so I switched to `platform.system()` for adding Linux support as well.


KillianLucas:
Thanks for testing this on Linux and pushing a fix! Let me know if there's anything else that doesn't work on Linux but should. Merging!

PiyushDuggal-source:
Sure, will be checking !

PR #1388, opened 2024-08-09
Ollama model list fix and a typo

CyanideByte:
- Filtered out ollama model list lines that aren't models ("failed", "NAME")
- Fixed a filename typo in the llamafile list

[?] Select a model:
   failed
   NAME
 > llama3.1:8b

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


KillianLucas:
Great catches @CyanideByte, merging!

Side note, I wonder if we should make it use the old `ollama` in a subprocess method if api_base isn't set, because I think for some things (like downloading a model) it shows a progress bar, wheras the API method wont. I might be wrong about that. Lmk if you have any thoughts

PR #1391, opened 2024-08-10
0.3.7 better server

KillianLucas:
Fixed some hallucination junk, made the server act better when it disconnects


interpreter/core/async_core.py
```diff
@@ -12,6 +12,7 @@
 
 import shortuuid
 from pydantic import BaseModel
+from starlette.websockets import WebSocketState
```
Notnaton, line 15: No starlette in pyproject.toml

Notnaton: https://github.com/OpenInterpreter/open-interpreter/pull/1411

PR #1445, opened 2024-09-06
Add sudo support for apt package installation in Terminal

lolsZz:
1. A new sudo_install method in the Terminal class that:                        
    • First attempts to install a package without sudo                          
    • If that fails, it prompts the user for their sudo password                
    • Then attempts to install the package using sudo with the provided password
 2. Modification of the run method to intercept apt install commands and use the 
   new sudo_install method                                                      

These changes address the issue where Open Interpreter would crash when trying  
to install packages that require sudo privileges. Now, it will handle such      
installations more gracefully, improving the user experience and expanding the  
capabilities of the interpreter.                                                

Key benefits:                                                                   

 • Allows for seamless installation of packages requiring sudo privileges       
 • Improves security by using a more secure method of passing the sudo password 
 • Provides user feedback on the installation process                           
 • Maintains the ability to install packages without sudo when possible         

Note: 
This implementation assumes the system is using apt as the package        
manager. Further modifications may be needed to support other package managers  
in the future. 

- [ * ] I have included relevant documentation updates (stored in /docs)
- [ * ] I have read `docs/CONTRIBUTING.md`
- [ * ] I have read `docs/ROADMAP.md`

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ * ] Tested on Linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -235,3 +235,4 @@ misc/
 
 # Ignore litellm_uuid.txt
 litellm_uuid.txt
+.aider*
--- a/interpreter/core/computer/terminal/terminal.py
+++ b/interpreter/core/computer/terminal/terminal.py
@@ -1,6 +1,8 @@
 import json
 import os
 import time
+import subprocess
+import getpass
 
 from ..utils.recipient_utils import parse_for_recipient
 from .languages.applescript import AppleScript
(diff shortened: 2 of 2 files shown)


KillianLucas:
Thank you so much @lolsZz! This is brilliant, super simple. I'm going to change it to be when the LLM decides to run sudo though, for user interpretability. (Seeing that it tries to run something that doesnt require sudo, then being prompted for password, I think might freak some people out)

Also, the LLM will try to install things with sudo, but because no input is possible, it fails. I love just checking it. We could have this trigger when it just starts with `sudo ` then it would work for everything.

PR #1533, opened 2024-11-15
Code execution never returns

guiramos:
I am trying to have my python project execute code dynamically via OI and every time it tries to run some shell script as described below, it never comes back. It gets stuck inside the method for line in interpreter.computer.run(language, code, stream=True):

This is the description generated.

First, I'll use SSH to connect to the server and run the disk usage command:
Code:
ssh -i /root/blabla.pem root@192.168.1.53 "df -h"
Filesystem Size Used Avail Use% Mounted on
tmpfs 778M 2.0M 776M 1% /run
/dev/sda2 440G 37G 380G 9% /
tmpfs 3.8G 0 3.8G 0% /dev/shm
tmpfs 5.0M 0 5.0M 0% /run/lock
tmpfs 778M 4.0K 778M 1% /run/user/1000

I think it gets stuck inside this loop because self.done never gets set.

        while True:
            if not self.output_queue.empty():
                yield self.output_queue.get()
            else:
                time.sleep(0.1)
            try:
                output = self.output_queue.get(timeout=0.3)  # Waits for 0.3 seconds
                yield output
            except queue.Empty:
                if self.done.is_set():
                    # Try to yank 3 more times from it... maybe there's something in there...
                    # (I don't know if this actually helps. Maybe we just need to yank 1 more time)
                    for _ in range(3):
                        if not self.output_queue.empty():
                            yield self.output_queue.get()
                        time.sleep(0.2)
                    break
                    
I am using version open-interpreter = "^0.4.3"

Can you please advise?

Reproduce
I am something like this:

for chunk in interpreter.chat(
f"{task_description}\n\n",
stream=True,
display=False
):

print out the chunks
Expected behavior
I would expect it to return, get out of the method.

Screenshots
No response

Open Interpreter version
0.4.3

Python version
3.11.5

Operating System name and version
mac 14

Additional context

### DIFF
--- a/interpreter/core/computer/terminal/languages/shell.py
+++ b/interpreter/core/computer/terminal/languages/shell.py
@@ -1,39 +1,10 @@
 import os
-import platform
+import queue
 import re
+import subprocess
+import threading
 
 from .subprocess_language import SubprocessLanguage
-
-
-class Shell(SubprocessLanguage):
-    file_extension = "sh"
-    name = "Shell"
-    aliases = ["bash", "sh", "zsh", "batch", "bat"]
-
-    def __init__(
-        self,
-    ):
-        super().__init__()
-
-        # Determine the start command based on the platform
-        if platform.system() == "Windows":
-            self.start_cmd = ["cmd.exe"]
-        else:
-            self.start_cmd = [os.environ.get("SHELL", "bash")]
-
-    def preprocess_code(self, code):
-        return preprocess_shell(code)
-
-    def line_postprocessor(self, line):
-        return line
-
-    def detect_active_line(self, line):
-        if "##active_line" in line:
-            return int(line.split("##active_line")[1].split("##")[0])
-        return None
-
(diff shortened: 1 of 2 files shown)


endolith:
I think I have the same problem, where using OI through SSH on a linux machine cannot run shell commands but can run python code.  When I try your change it does seem to fix this problem, but it seems to break running commands on windows, they return nothing?

```shell

▌ Model set to gpt-4o-mini

Press CTRL-C to exit.

> ls


  ls


  Would you like to run this code? (y/n)

  y


  ls


  ""


  It seems that the ls command did not return any files or directories, which could mean the current directory is empty. Would you like to create a file, navigate somewhere, or perform another action?

>
```

endolith:
(Nevermind.  The reason OI wasn't able to run commands in Linux any longer was because I use `fish` as my default shell.  Previously I had set `conda env config vars set SHELL=/bin/bash` for the environment, but I must have recreated the environment and forgotten that step.)

PR #1583, opened 2025-01-26
Improve truncation

endolith:

When truncating long outputs, remove the middle and show the head and tail to the assistant, instead of showing only the tail.  This provides more useful context.  

Also improve the suggestions given to the assistant for how to deal with long output, since it tends to do the wrong thing here.

I don't see any.

- [x] I have read `docs/CONTRIBUTING.md`  [but running `black` and `isort` on the code caused many unrelated changes, so I didn't commit that.]

- [x] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/interpreter/core/utils/truncate_output.py
+++ b/interpreter/core/utils/truncate_output.py
@@ -9,8 +9,11 @@ def truncate_output(data, max_output_chars=2800, add_scrollbars=False):
 
     message = ("Output truncated. "
                f"Showing {chars_per_end} characters from start/end. "
-               "You should try again and use computer.ai.summarize(output) "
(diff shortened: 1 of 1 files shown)


endolith:
Also prints the total length now, with typical output like this:

```
…
      print(content)


  Output truncated (2,659,274 characters total). Showing 1,400 characters from start/end. To handle large outputs,
  store result in python var first `result = command()` then `computer.ai.summarize(result)` for a summary, search
  with `result.find('text')`, repeat shell commands with wc/grep/sed, etc. or break it down into smaller steps.

  <?xml version='1.0' encoding='UTF-8' standalone='yes' ?>
…
```

PR #1586, opened 2025-01-31
Include x-coding-assistant=open-interpreter header in litellm calls

jhrozek:

Proxies that inspect traffic between the development environment and an
LLM might be interested in whether it's OI or another tool calling in
order to be able to inspect and/or modify the payload.

The most common way of solving this would be to add a `user-agent` header.
However, litellm which OI uses calls into OpenAI libraries directly
when making the request and it seems like the only way to set a custom
`http_client`. This seemed like something that might have unforeseen
consequences (timeouts? retries?). For other LLMs, litellm seems to use
its own httpx wrapper which might presumably be easier to customize, but
I have not tried.

To make things easier, let's just add an OI specific header. I put the
string OI followed by the version there, but the value - and indeed the key
- of the header are not that interesting, what I would like to do is to just
be able to to tell OI calls.

N/A

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/llm/__init__.py
+++ b/interpreter/core/llm/__init__.py
@@ -1,0 +1,14 @@
+from importlib import metadata
+
+_VERSION = "dev"
+
+
+def __get_version():
+    try:
+        version = metadata.version("open-interpreter")
+    except metadata.PackageNotFoundError:
+        version = _VERSION
+    return version
+
+
+__version__ = __get_version()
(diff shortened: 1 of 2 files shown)


jhrozek:
@KillianLucas is this something that OpenInterpreter would be interested in merging?

PR #1609, opened 2025-03-25
Don't check equality with None

endolith:

Replace `== None` with `is None` 

https://www.flake8rules.com/rules/E711.html

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/interpreter.py
+++ b/interpreter/interpreter.py
@@ -359,7 +359,7 @@ class Interpreter:
                             current_block.partial_json += chunk.delta.partial_json
 
                             if hasattr(current_block, "name"):
-                                if edit.name == None:
+                                if edit.name is None:
                                     edit.name = current_block.name
                                 edit.feed(chunk.delta.partial_json)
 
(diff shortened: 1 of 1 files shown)


endolith:
(This section is pretty hard to read.  Would be better to give meaningful names to things like `chunk.choices[0].delta.content`?)

Notnaton:
> > > (This section is pretty hard to read. Would be better to give meaningful names to things like `chunk.choices[0].delta.content`?)
> > 
> > 
> > No, this is the official API https://docs.litellm.ai/docs/
> 
> I don't see it there?
> 
> I mean things like, instead of
> 
> ```python
> if chunk.choices[0].delta.content:
>     md.feed(chunk.choices[0].delta.content)
>     await asyncio.sleep(0)
> 
>     if message.content is None:
>         message.content = chunk.choices[0].delta.content
>     elif chunk.choices[0].delta.content is not None:
>         message.content += chunk.choices[0].delta.content
> ```
> 
> it could be:
> 
> ```python
> delta = chunk.choices[0].delta
> if delta.content:
>     md.feed(delta.content)
>     await asyncio.sleep(0)
> 
>     if message.content is None:
>         message.content = delta.content
>     elif delta.content is not None:
>         message.content += delta.content
> ```
> 
> or other simplifications, instead of repeatedly accessing deeply nested things

Aah I see, I think that would be ok

PR #1638, opened 2025-08-06
Update README.md

Notnaton:

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,15 +13,6 @@
     <br><a href="https://0ggfznkwh4j.typeform.com/to/G21i9lJ2">Get early access to the desktop app</a>‎ ‎ |‎ ‎ <a href="https://docs.openinterpreter.com/">Documentation</a><br>
 </p>
 
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


Notnaton:
Should we recommend usage of pipx/ uv tool install?

endolith:
I was thinking more of like "there are currently three versions of open Interpreter: main branch (status is...), Dev branch 1.0 (status is...) and oiv2 (status is...)"

endolith:
> Should we recommend usage of pipx/ uv tool install?

Doesn't matter to me. Whatever gives the smoothest experience. I used pipx before switching to pip install -e. Maybe describe both?

endolith:
@Notnaton Should the Quick start section also install from github instead of pypi?  Are you not able to update pypi? https://github.com/openinterpreter/open-interpreter/blob/main/README.md#quick-start

Notnaton:
Yes it should, good catch.

PR #1644, opened 2025-09-03
feat: plumb `reasoning_effort` through Open Interpreter → LiteLLM

M-J-W1:
Adds a `reasoning_effort` parameter ("minimal" | "low" | "medium" | "high") and passes it through to LiteLLM for reasoning-capable models.

Changed files:
- interpreter/core/core.py
- interpreter/core/llm/llm.py

Usage:
interpreter.llm.reasoning_effort = "minimal"

External example: https://github.com/uhsealevelcenter/IDEA/blob/main/app.py

Backwards compatibility: default behavior unchanged if not set.


interpreterwork:
Thank you for contributing this reasoning-effort support. This pull request targets the former Python/LiteLLM implementation on the old `development` branch.

Open Interpreter `main` is now a new Rust implementation based on Codex, with a different provider, model, and reasoning architecture. The original Python project continues as the community-maintained [`endolith/open-interpreter`](https://github.com/endolith/open-interpreter) fork, which is the appropriate place to revive this Python change if it is still useful.

If you are interested in the current project instead, please try the Rust/Codex-based version on `main`; any current reasoning-control change would need to be designed against that implementation. Closing this legacy-targeted PR here, with thanks for the work.

PR #1682, opened 2026-01-23
fix: R function return values return NULL instead of actual values (#1655)

### LINKED ISSUE #1655: R Function Return Values Return NULL Instead of Actual Values
## Description

When running R code in Open Interpreter that defines and calls functions, the function's return value is often captured as `NULL` instead of the actual computed value. Simple print statements and direct assignments work fine, but function calls often fail to capture return values properly.

### What Works
- Simple prints: `print("Hello, world!")` ✓
- Direct calculations: `normalized_x <- (x - min(x)) / (max(x) - min(x))` ✓

### What Fails
- Function return values show as `NULL` when using implicit returns

## Technical Details

The preprocessing in `interpreter/core/computer/terminal/languages/r.py` wraps all code in `tryCatch()` blocks (line 32-38), which suppresses function return values. The `tryCatch` wrapper is intended for error handling but ends up suppressing normal R function returns.

## Suggested Fix

Modify the `tryCatch()` wrapper in `r.py` to preserve and display return values or use a different error handling approach that doesn't suppress returns.

## Steps to Reproduce

1. Start an Open Interpreter session
2. Ask the AI to create a normalization function in R, for example: 
````r
normalize <- function(x) {
    (x - min(x)) / (max(x) - min(x))
}
result <- normalize(c(1, 2, 3, 4, 5))
print(result)
````
3. Observe output shows `NULL` instead of expected normalized values

## Expected Behavior

Function return values should be captured and displayed correctly, regardless of whether they use implicit or explicit returns.

## Actual Behavior

Function return values are captured as `NULL` when using implicit returns (which is the normal R convention - the value of the last evaluated expression).

### Screenshots

_No response_

### Open Interpreter version

0.4.3 (develop branch)

### Python version

3.10.13

Windows 10

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by majiayu000**

When R code with functions using implicit returns (standard R convention) was executed, the return value was NULL instead of the actual computed value. This was caused by active line markers being injected before closing braces/brackets/parens, making `cat()` the last expression in blocks instead of the actual return value.

**Root Cause:**
The preprocessing code was adding `cat("##active_lineN##\n");` before EVERY line, including structural closing lines like `}`. In R, `tryCatch` returns the value of the last expression, and when that expression is `cat()` (which returns NULL invisibly), the implicit return becomes NULL.

**Changes:**
- Skip injecting active line markers on non-executable lines:
  - Blank lines
  - Comment lines (starting with #)
  - Lines starting with closing tokens: `)`, `]`, `}`
- Add comprehensive unit tests for R preprocessing

**Example that now works correctly:**
```r
normalize <- function(x) {
    (x - min(x)) / (max(x) - min(x))
}
result <- normalize(c(1, 2, 3, 4, 5))
print(result)
# Previously: printed NULL
# Now: prints [1] 0.00 0.25 0.50 0.75 1.00
```

Fixes #1655

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/interpreter/core/computer/terminal/languages/r.py
+++ b/interpreter/core/computer/terminal/languages/r.py
@@ -22,7 +22,21 @@ class R(SubprocessLanguage):
         processed_lines = []
 
         for i, line in enumerate(lines, 1):
-            # Add active line print
+            stripped = line.strip()
+            lstripped = line.lstrip()
+
+            # Skip injecting markers on non-executable structural lines.
+            # Otherwise the trailing cat() becomes the last expression in a block
+            # and implicit returns become NULL instead of the actual value.
+            # This affects blank lines, comments, and closing tokens like }, ), ]
+            if not stripped or lstripped.startswith("#"):
+                processed_lines.append(line)
+                continue
+            if lstripped and lstripped[0] in ")]}":
+                processed_lines.append(line)
+                continue
+
+            # Add active line print for executable lines
             processed_lines.append(f'cat("##active_line{i}##\\n");{line}')
 
         # Join lines to form the processed code
--- a/tests/core/computer/test_r_language.py
+++ b/tests/core/computer/test_r_language.py
@@ -1,0 +1,106 @@
+"""Tests for R language preprocessing in interpreter."""
+
+import unittest
+
+from interpreter.core.computer.terminal.languages.r import R
+
+
+class TestRPreprocessing(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

endolith:
There is no need to close the PR if it's valid

PR #1691, opened 2026-02-19
Harden macOS hotkey execution against command injection

lawyered0:
## Summary
- replace shell-based `os.system("osascript -e ...")` invocation with argument-safe `subprocess.run(["osascript", "-e", script], ...)` in both macOS hotkey paths
- normalize and allowlist modifier tokens for AppleScript hotkeys, with fallback to existing `pyautogui.hotkey(...)` behavior when invalid
- escape keystroke content before embedding into AppleScript to prevent script-breaking/injection via quotes and backslashes

## Security impact
Previously, hotkey text could be interpolated into a shell command string. A crafted payload containing a single quote could break out of the quoted shell argument and execute arbitrary shell commands.

This patch removes shell invocation from these paths and hardens AppleScript construction.

## Tests added
- `tests/core/computer/test_keyboard_hotkey_security.py`
  - macOS path uses subprocess list-args
  - escaping of dangerous keystroke content
  - invalid-modifier fallback
  - non-macOS fallback behavior
- `tests/computer_use/tools/test_computer_hotkey_security.py`
  - `ComputerTool` macOS key action uses `_run_macos_hotkey`
  - fallback to `pyautogui.hotkey` when hotkey runner fails
  - `_run_macos_hotkey` uses subprocess list-args and escapes content
  - invalid modifier is rejected without subprocess execution

## Validation run
- `.venv/bin/python -m pytest -q tests/computer_use/tools/test_computer_hotkey_security.py tests/core/computer/test_keyboard_hotkey_security.py tests/core/test_async_core.py`
- `.venv/bin/python -m compileall -q interpreter/core/computer/keyboard/keyboard.py interpreter/computer_use/tools/computer.py tests/core/computer/test_keyboard_hotkey_security.py tests/computer_use/tools/test_computer_hotkey_security.py`

Both completed successfully locally.

## Backward compatibility
No public API signatures changed. This is a behavior hardening change for macOS hotkey internals.

### DIFF
--- a/interpreter/computer_use/tools/computer.py
+++ b/interpreter/computer_use/tools/computer.py
@@ -1,10 +1,8 @@
 import asyncio
 import base64
 import math
-import os
 import platform
-import shlex
-import shutil
+import subprocess
 import tempfile
 import time
 from enum import StrEnum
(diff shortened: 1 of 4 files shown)


regaan:
Hi @lawyered0, thanks for the fix! I actually reported these exact issues privately on Feb 5 (GHSA-q9r3-mjf7-pmw9). Since the maintainers haven't responded to the private report or this PR, I am proceeding with MITRE for the CVE assignment to ensure this gets a proper security identifier. Thanks for helping secure the repo!

PR #1698, opened 2026-02-25
fix: replace 55 bare except clauses with except Exception

haosenwang1018:
## What
Replace 55 bare `except:` clauses with `except Exception:`.

## Why
Bare `except:` catches `BaseException`, including `KeyboardInterrupt` and `SystemExit`, which can prevent clean process shutdown and mask critical errors. Using `except Exception:` catches all application-level errors while allowing system-level exceptions to propagate correctly.

### DIFF
--- a/interpreter/core/archived_server_2.py
+++ b/interpreter/core/archived_server_2.py
@@ -90,7 +90,7 @@ class AsyncInterpreter:
         else:
             try:
                 chunk = json.loads(chunk)
-            except:
+            except Exception:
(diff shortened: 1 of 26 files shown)


endolith:
Does this just replace all such cases or does it consider the ones where it's intended to catch KeyboardException?

ffbb90bb fb4b8caf are related

haosenwang1018:
Closing per review backlog policy; can resubmit in smaller chunks.

PR #1707, opened 2026-03-13
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade MiniMax model configuration to use M3 as the default model.

## Changes
- Set `MiniMax-M3` as the default model in the MiniMax profile
- Keep `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as alternatives (older M2.5 variants removed)
- Update context window to 512K and enable vision support
- Update README documentation to reflect new default model

## Why
MiniMax-M3 is the latest flagship model, with a 512K context window, up to 128K output, and image input support.

## Testing
- Profile loads successfully with M3 set as default
- MINIMAX_API_KEY environment variable is honored

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,32 @@ interpreter.llm.model = "gpt-3.5-turbo"
 
 [Find the appropriate "model" string for your language model here.](https://docs.litellm.ai/docs/providers/)
 
+### Using MiniMax
+
+To use [MiniMax](https://platform.minimax.io/) models, set your API key and use the built-in profile:
+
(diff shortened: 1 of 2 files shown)


endolith:
I have a fork with support for open router and reasoning tokens by the way

octo-patch:
Thanks for sharing, @endolith! OpenRouter and reasoning token support sounds great. If this PR gets traction, it would be nice to see those capabilities integrated too.

PR #1709, opened 2026-03-14
Preserve truncated output recovery details

### LINKED ISSUE #1695: Bug: truncated command output permanently lost, message references non-existent method
## Description

Two issues with `truncate_output()` in `interpreter/core/utils/truncate_output.py`:

1. **Output permanently lost**: When command output exceeds `max_output_chars`, the middle portion is replaced with `[...]` and discarded. Neither the user nor the LLM can recover the full output. For commands like `ls -la` on a large directory or `grep -r` results, the most useful data is often in the discarded portion.

2. **Truncation message references non-existent method**: The message tells users to use `computer.ai.summarize(result)`, which does not exist in the CLI context. Users follow this suggestion and get an error.

3. **ANSI escape codes counted as content**: Raw ANSI codes are counted toward the character limit, meaning the visible content is shorter than `max_output_chars` suggests.

## Current code

```python
message = (f"Output truncated ({len(data):,} characters total). "
           f"Showing {chars_per_end:,} characters from start/end. "
           "To handle large outputs, store result in python var first "
           "`result = command()` then `computer.ai.summarize(result)` for "
           "a summary, search with `result.find('text')`...")
```

## Suggested fix

- Strip ANSI escape codes before measuring length
- Save full output to a temp file (e.g. `/tmp/oi-output-latest.txt`) before truncating, so users can review it
- Reword the truncation message to reference practical shell commands (`head`, `tail`, `grep`) instead of non-existent Python methods

Related: #1588

## Environment

- open-interpreter 0.4.3
- Python 3.10, Linux

biefan:
## Summary
- measure truncation based on visible characters so ANSI escape sequences do not trigger premature truncation
- save the full output to a temp file before truncating it
- replace the broken `computer.ai.summarize(result)` advice with shell commands users can actually run
- add regression tests for ANSI-heavy output and temp-file recovery

## Testing
- `uv run --python 3.11 --with pytest python -m pytest tests/core/utils/test_truncate_output.py -q`
- `python3 -m py_compile interpreter/core/utils/truncate_output.py tests/core/utils/test_truncate_output.py`

Closes #1695

### DIFF
--- a/interpreter/core/utils/truncate_output.py
+++ b/interpreter/core/utils/truncate_output.py
@@ -1,3 +1,30 @@
+import re
+import tempfile
+from pathlib import Path
+
+
+ANSI_ESCAPE_RE = re.compile(r"\x1B\[[0-?]*[ -/]*[@-~]")
+TRUNCATION_PREFIX = "Output truncated ("
+
+
+def _strip_ansi_codes(data):
+    return ANSI_ESCAPE_RE.sub("", data)
+
+
+def _latest_output_path():
+    return Path(tempfile.gettempdir()) / "oi-output-latest.txt"
+
+
+def _build_truncation_message(total_chars, chars_per_end, output_path):
+    return (
+        f"Output truncated ({total_chars:,} visible characters total). "
+        f"Showing {chars_per_end:,} visible characters from start/end. "
+        f"Full output saved to `{output_path}`. "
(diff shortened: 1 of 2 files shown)


endolith:
Hmm, but truncated output applies to python execution and other languages, too, so is it good to handle that output in shell?  Or should it be saved as a python object?

PR #1729, opened 2026-04-17
fix: detect sudo password prompt and exit cleanly instead of hanging

### LINKED ISSUE #1694: Bug: sudo commands hang process indefinitely with no feedback
## Description

When the LLM generates a command that triggers a `[sudo] password for <user>` prompt, the subprocess hangs indefinitely waiting for password input that can never arrive. OI uses `subprocess.Popen` with `stdin=subprocess.PIPE` — there is no terminal for password entry. The process blocks until the execution timeout fires (default 120s), with no indication of what went wrong.

## Steps to reproduce

1. Start OI with `auto_run = True` (or approve when prompted)
2. Ask it to do something that requires root, e.g. "check the systemd journal for errors"
3. The LLM generates `sudo journalctl -xe`
4. OI hangs for ~120 seconds with no output
5. Eventually times out with a generic error

## Root cause

`handle_stream_output()` in `subprocess_language.py` reads lines via `iter(stream.readline, "")` and checks for `KeyboardInterrupt`, active line markers, and end-of-execution markers — but has no detection for sudo password prompts. The `[sudo] password for` message arrives on stderr, gets queued as regular output, and the process continues blocking on stdin.

## Suggested fix

Detect the sudo prompt pattern in the output stream and immediately signal completion with a clear message:

```diff
  def handle_stream_output(self, stream, is_error_stream):
      try:
          for line in iter(stream.readline, ""):
              ...
              line = self.line_postprocessor(line)

              if line is None:
                  continue

+             if "[sudo] password for" in line:
+                 self.output_queue.put(
+                     {
+                         "type": "console",
+                         "format": "output",
+                         "content": "[sudo detected — command requires elevated "
+                                    "privileges. Run it manually.]\n",
+                     }
+                 )
+                 self.done.set()
+                 return

              if self.detect_active_line(line):
                  ...
```

Since the subprocess uses pipes (not a PTY), we cannot send Ctrl+C to cancel the sudo prompt. But `self.done.set()` causes `run()` to stop waiting and return. The blocked subprocess is cleaned up when `terminate()` is called on the next execution or OI exit.

## Environment

- open-interpreter 0.4.3
- Python 3.10, Linux (ARM64)
- Ollama backend, bash shell execution

octo-patch:
Fixes #1694

## Problem
When the LLM generates a `sudo` command (e.g., `sudo journalctl -xe`), the subprocess uses `subprocess.PIPE` which has no terminal for interactive password input. The `[sudo] password for <user>:` prompt appears on stderr but the process then blocks indefinitely waiting for input that can never arrive through the pipe. OI hangs for ~120 seconds with no output before timing out.

## Solution
Added a check in `handle_stream_output()` in `subprocess_language.py` for the `[sudo] password for` pattern in the output stream. When detected, OI immediately:
1. Emits a clear message informing the user that sudo requires elevated privileges and should be run manually
2. Sets the `done` event so the `run()` loop exits without waiting for the execution timeout

The subprocess remains blocked but gets cleaned up on the next execution or when OI exits, consistent with how other error cases are handled.

## Testing
- Trigger a sudo command with OI in a non-passwordless sudo environment
- Before: OI hangs for ~120 seconds then times out with a generic error
- After: OI immediately returns with a clear message: `[sudo] password prompt detected — OI cannot provide a password through a pipe. Please run this command manually in your terminal.`

Existing `True`/`False` auto_run behavior and non-sudo commands are completely unaffected.

### DIFF
--- a/interpreter/core/computer/terminal/languages/subprocess_language.py
+++ b/interpreter/core/computer/terminal/languages/subprocess_language.py
@@ -148,6 +148,20 @@ class SubprocessLanguage(BaseLanguage):
                 if line is None:
                     continue  # `line = None` is the postprocessor's signal to discard completely
 
+                if "[sudo] password for" in line:
+                    self.output_queue.put(
+                        {
+                            "type": "console",
+                            "format": "output",
+                            "content": (
+                                "[sudo] password prompt detected — OI cannot provide a password "
+                                "through a pipe. Please run this command manually in your terminal.\n"
+                            ),
+                        }
+                    )
+                    self.done.set()
+                    return
+
                 if self.detect_active_line(line):
                     active_line = self.detect_active_line(line)
                     self.output_queue.put(


endolith:
It doesn't actually block, you can type the password and it runs. I usually just press enter so it says incorrect password and then type the correct password so I know it's going into the right place. Definitely should be handled better though

octo-patch:
Thanks for the context! You're right that you can type the password and it will work — the blocking behavior just means there's no visible prompt to indicate what's happening, which can confuse users who don't realize sudo is waiting. This PR aims to make that situation explicit (detect and exit cleanly with a clear message) so the user knows to either run their command without sudo or set up passwordless sudo. Happy to revisit the approach if you have a different preference for handling this UX issue.

endolith:
I wonder if there's a way to just show the password prompt?

Notnaton:
We could replace sudo with pkexec.
On gui distro it will pop up a gui.

This may break if run through ssh?

endolith:
Actually Cursor was able to fix this in my fork by changing one line: https://github.com/endolith/open-interpreter/commit/2e43fec9af126c9912e7e601f36c4c7bc6f2151f

Not sure if that has other side effects or if it only works in my fork vs in the original, though.

see also https://github.com/openinterpreter/open-interpreter/issues/1694

PR #1738, opened 2026-04-25
Make basic usage look fresher

mikermcneil:
If I’m wrong about this and claude2 is suggested for a reason, rather than sonnet, might be helpful to alternatively leave a comment in the code sample explaining why.


endolith:
The repo has not been updated in a long time and many of the models are outdated and should be updated

PR #1765, opened 2026-05-25
Enhancement : add per-response token usage display in terminal

### LINKED ISSUE #1754: [Enhancement] Show token count and cost estimate after each assistant message
## Description

Open Interpreter runs code locally with full internet access, but provides no token usage visibility. Unlike cloud-based code interpreters (ChatGPT Code Interpreter, Claude Code), users have no way to see how many tokens each assistant message consumed or estimate the cost.

OI already uses readline for history (per issue #1697). Adding token/cost tracking would complement that infrastructure.

## Suggested implementation

Track tokens per assistant response and display after each message:
- Add a %tokens on/off magic command to toggle (persists in session)
- Show always-on when --verbose flag is set
- Add --track-cost CLI flag

## Use Case

Developers testing different models and budget-conscious users want to monitor token usage and estimated costs in real-time, especially when iterating on prompts or processing large datasets.

## Additional Information

This could be gated behind a config flag track_usage: true in ~/.config/open-interpreter/config.yaml for users who want it enabled by default.

kuldeeepy:
## Add per-response token usage display in the terminal

  This PR implements issue #1754 by showing token usage after each completed assistant response in the terminal.

Screenshot 2026-05-25 at 10 59 17 PM

  ### What changed

  - Added terminal usage display after completed assistant responses
  - Added `%tokens` as a session toggle for usage display
  - Added `--track-cost` / `--track_cost` to enable usage display at startup
  - Made `--verbose` also enable usage display for that run
  - Preserved the old next-request estimator under `%token_estimate`
  - Added a fallback estimator for providers/local models that do not return usage metadata
  - Marked fallback output clearly as estimated: `Tokens (estimated)`

  ### Notes

  - Provider-reported usage is preferred when available
  - Local models like `phi3` may not return usage metadata, so this PR falls back to estimating prompt/completion tokens from the actual
  request/response content
  - Usage display is emitted only after a completed assistant turn

  ### Testing

  Tested from the source tree with a local model:

  - Launch:
    `PYTHONPATH=. ./ven/bin/python -c "from interpreter.terminal_interface.start_terminal_interface import main; main()" --local`
  - Enable tracking:
    `%tokens`
  - Verified completed assistant turns display:
    `> Tokens (estimated): ...`

  Also verified:

  - `%tokens` toggles tracking on/off
  - the source-tree CLI parser accepts `--track-cost`
  - fallback estimation works when provider usage metadata is unavailable

  Fixes #1754

  - [ ] Tested on Windows
  - [x] Tested on MacOS
  - [ ] Tested on Linux

### DIFF
--- a/interpreter/core/core.py
+++ b/interpreter/core/core.py
@@ -100,6 +100,9 @@ class OpenInterpreter:
         self.contribute_conversation = contribute_conversation
         self.plain_text_display = plain_text_display
         self.highlight_active_line = True  # additional setting to toggle active line highlighting. Defaults to True
+        self.track_usage = False
+        self.last_usage = None
+        self.last_usage_request = None
 
         # Loop messages
         self.loop = loop
@@ -310,6 +313,8 @@ class OpenInterpreter:
                 return True
             if chunk["type"] == "review":
                 return True
+            if chunk["type"] == "usage":
+                return True
             return False
 
         last_flag_base = None
@@ -432,6 +437,8 @@ class OpenInterpreter:
         self.computer._has_imported_computer_api = False  # Flag reset
         self.messages = []
         self.last_messages_count = 0
+        self.last_usage = None
+        self.last_usage_request = None
 
     def display_message(self, markdown):
         # This is just handy for start_script in profiles.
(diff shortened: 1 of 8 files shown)


kuldeeepy:
**One implementation note:** 

> when the provider returns real usage metadata, the terminal displays that. When it does not (eg with  local models), this falls back to estimated prompt/completion tokens using LiteLLM token counting.

endolith:
I have a  `%usage` command here which is similar: https://github.com/openinterpreter/open-interpreter/commit/409f5edd

`````
> %usage


  ▌ Last model response — usage from the HTTP call that produced the assistant message you just saw.

  ▌ For normal chat that is one call per reply. If the assistant used tools (e.g. code execution), Open Interpreter may call the model more than once before you see the reply; %usage
  ▌ then shows only the latest of those calls, not a sum.

  ▌ Reading values: not reported means the API sent JSON null or omitted meaning for that field. A numeric 0 means the API explicitly reported zero (e.g. zero cache-hit tokens).

   • Prompt tokens: 58590
   • Completion tokens: 391
   • Total tokens: 58981
   • Completion — Accepted Prediction Tokens: not reported
   • Completion — Audio Tokens: not reported
   • Completion — Image Tokens: not reported
   • Completion — Reasoning Tokens: 559
   • Completion — Rejected Prediction Tokens: not reported
   • Completion — Text Tokens: not reported
`````

PR #1778, opened 2026-06-04
fix: route "cmd" code blocks to the shell instead of rejecting them

### LINKED ISSUE #1023: "cmd" disabled or not supported
### Describe the bug

image

### Reproduce

The language model is: bagelmisterytour
not really relevant

### Expected behavior

We should make it use powershell when cmd is the input from the model

### Screenshots

_No response_

### Open Interpreter version

0.2.0

### Python version

n/a

Windows 11

### Additional context

_No response_

koriyoshi2041:
Models routinely tag Windows command-prompt code blocks as `cmd`, but that language name was never registered, so it fell through to "`cmd` disabled or not supported." and the code was silently dropped.

Root cause: the Shell language lists `batch`/`bat`/`bash`/etc. as aliases but not `cmd`, so `Terminal.get_language("cmd")` returns `None`. The Shell language already launches `cmd.exe` on Windows, so it can run this code — it just wasn't reachable by that name.

Fix: add `"cmd"` to the Shell aliases. `cmd` blocks now resolve to the shell and run like `bat`/`batch` already do.

Added tests/core/computer/test_terminal.py covering `get_language("cmd")`, case-insensitivity, and the existing aliases as a guard. `pytest tests/core/computer/` passes (8 tests); black/isort clean.

Fixes #1023

### DIFF
--- a/interpreter/core/computer/terminal/languages/shell.py
+++ b/interpreter/core/computer/terminal/languages/shell.py
@@ -8,7 +8,7 @@ from .subprocess_language import SubprocessLanguage
 class Shell(SubprocessLanguage):
     file_extension = "sh"
     name = "Shell"
-    aliases = ["bash", "sh", "zsh", "batch", "bat"]
+    aliases = ["bash", "sh", "zsh", "batch", "bat", "cmd"]
 
     def __init__(
         self,
--- a/tests/core/computer/test_terminal.py
+++ b/tests/core/computer/test_terminal.py
@@ -1,0 +1,27 @@
+import unittest
+from unittest import mock
+
+from interpreter.core.computer.terminal.languages.shell import Shell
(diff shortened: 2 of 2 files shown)


endolith:
In my fork I got rid of `shell` entirely and the LLM uses `cmd` `bash` or `powershell` directly.  It seems to work much better, without confusing the syntaxes of one while using the other.  (And it can choose between `bash` (from Git for Windows) or `cmd` depending on what works better for a particular task.)

koriyoshi2041:
That direction makes sense to me. I kept this PR narrower because the current code still exposes `shell` as the shell-execution language and was rejecting `cmd` before it reached that path.

If the project wants to remove `shell` entirely and have the model choose `cmd` / `bash` / `powershell` directly, I agree that would be cleaner, but it feels like a separate compatibility change from this small Windows fix.

PR #1795, opened 2026-07-12
Fix SWE-agent link in harness documentation

ofirpress:
Updated the SWE-agent link to point to the correct website.

I'm one of the team members of the SWE-agent project, and that domain was never used by us. I fixed it. Thanks


KillianLucas:
Thanks Ofir!

PR #1797, opened 2026-07-13
fix: hide models requiring a newer Codex client

### LINKED ISSUE #1787: Requires newer version of Codex.

I have just installed OpenInterpreter, and when I try to send messages, it shows an error:
```
■ {"type":"error","status":400,"error":{"type":"invalid_request_error","message":"The 'gpt-5.5' model requires a newer
version of Codex. Please upgrade to the latest app or CLI and try again."}}
```

Start the program and send a random text

It should be working and chatting?

### Additional information

_No response_

ferminquant:
## What Problem This Solves

Open Interpreter 0.0.21 can expose a catalog model that requires a newer Codex client, allowing it to be selected as the default and fail at runtime with the issue's compatibility error.

## Why This Change Was Made

The model catalog already provides `minimal_client_version`, but `ModelInfo` discarded it. This change preserves the metadata, accepts both the array form returned by the endpoint and the string form in the bundled catalog, and removes incompatible models before picker/default selection.

## User Impact

Models requiring a newer client are no longer offered by the model picker. Compatible models, including `gpt-5.2`, remain available.

## Evidence

- Before the fix, the new regression test reproduced the failure by selecting the incompatible model as default.
- `just test -p codex-models-manager`: 49 passed
- `just test -p codex-protocol`: 243 passed
- `cargo check --workspace`: passed

Fixes #1787


KillianLucas:
Thanks @ferminquant! 

By merging upstream Codex this should be fixed. Let me know if its not!

PR #1819, opened 2026-07-18
Complete Kimi Code request continuity

KillianLucas:
## What this fixes

Kimi sessions were missing parts of the Kimi Code tool surface and could lose provider-required reasoning context between tool calls, assistant replies, and resumed follow-up turns. File, media, planning, task, and background-agent behavior also differed from the provider's expected semantics, while CRLF checkouts on Windows could leave prompt-template markers visible.

## Changes

- order project skills before provider built-ins, deduplicate them by name, and cap descriptions safely
- preserve Kimi reasoning across tool turns and combine narration, reasoning, and tool calls correctly
- support Kimi image and video input through the native media tool
- implement Kimi-compatible Glob, Grep, Read, Write, Edit, and Bash behavior
- support todo progress reminders and native plan-mode tools
- support native background process tasks, task inspection, task stopping, and background agents
- support native Agent and AgentSwarm execution without invoking an external agent CLI
- normalize embedded prompts and skills across platforms
- retain entirely native Rust execution

## Validation

- regression coverage for skill ordering, deduplication, path preference, and truncation
- regression coverage for durable reasoning and combined assistant/tool messages
- regression coverage for media, recursive Glob, Grep output modes, Read limits, Write/Edit behavior, Bash output shaping, todo reminders, and task state
- Windows path and line-ending regression coverage
- `just fmt`
- `git diff --check`
- clean hosted GitHub checks


interpreterwork:
Closing this stale integration branch rather than merging it into the current product.

The branch is now more than two hundred `main` commits behind and adds over three thousand lines against an older harness/runtime shape. Since it was opened, Open Interpreter has completed a large stable Codex sync and the Kimi integration has continued evolving on the new architecture. Porting this branch wholesale would reintroduce obsolete assumptions and make the current delta substantially harder to maintain.

The product goals remain valid. Any Kimi behavior still missing on current `main` should return as focused, current-architecture changes with current regression coverage rather than through this historical umbrella branch.

PR #1823, opened 2026-07-18
fix(rmcp-client): cap streamable HTTP response body size

ethan0905:
### What I found

While auditing the MCP client transport, I noticed that `collect_body` in
`rmcp-client/src/http_client_adapter.rs` reads an HTTP response from an MCP
server into memory with **no upper bound**:

```rust
let mut body = Vec::new();
while let Some(chunk) = body_stream.recv().await... {
    body.extend_from_slice(&chunk);
}
```

Every chunk the server sends is appended to `body` until the stream ends. The
size is entirely controlled by the remote MCP server, and only *after* the whole
thing is buffered do we hand it to `serde_json::from_slice`.

What stood out is that the crate already knows this is a risk on a neighboring
code path: the OAuth client (`oauth_http_client.rs`) caps responses at 1 MiB via
`MAX_OAUTH_HTTP_RESPONSE_BODY_BYTES`. The main data path just never got the same
guard.

Any MCP server the user connects to (including a compromised or buggy one) can
reply to a routine request like `tools/list` with a multi-gigabyte or
never-ending chunked body. The client dutifully buffers all of it, so memory
grows without limit until the process is killed by the OS — a denial-of-service
that takes the whole interpreter session down. The user doesn't have to do
anything unusual; a single normal request to a bad server is enough.

### How I fixed it

- Enforce a **100 MiB** ceiling while buffering in `collect_body`. If a response
  would exceed it, we stop and return a new `ResponseBodyTooLarge` error instead
  of continuing to allocate.
- The bounds check uses saturating subtraction (`chunk_len <=
  limit.saturating_sub(current_len)`) so it can never underflow/panic.
- `ResponseBodyTooLarge` is classified as **non-retryable**, since retrying would
  just pull the same oversized body again.
- I pulled the boundary logic into a tiny pure helper (`within_body_limit`) so it
  can be unit-tested directly.

I picked 100 MiB deliberately: it's generous enough to never reject a legitimate
large `resources/read` or `tools/list`, while still bounding what a malicious
server can force us to allocate.

### Why this only makes the interpreter better

The cap is a ceiling, not a preallocation — we still grow the buffer chunk by
chunk and only hold the bytes actually received. So for every normal response
(kilobytes to a few MB) behavior, speed, and memory are **identical** to before;
the only new behavior is that a pathological response is rejected instead of
crashing the process. Nothing that worked before stops working, and the streaming
(SSE) path isn't touched at all. It's purely a safety net that follows a pattern
the codebase already established for OAuth.

### Scope / known remaining vector

This PR caps the **buffered** response path (`collect_body`), which handles
`application/json` and error bodies. It intentionally does **not** cover the
**SSE** path (`Content-Type: text/event-stream` → `sse_stream_from_body`), so a
server could still stream an unbounded single event to grow memory. I scoped it
out on purpose: the SSE buffering lives inside the third-party `sse-stream` crate
(no size-limit option), and SSE is a legitimately long-lived, high-volume stream,
so a naive cumulative cap would risk disconnecting valid sessions. Bounding it
correctly needs a per-event limit (an upstream `max_event_size` in `sse-stream`
or a carefully-tested bounding adapter) and is best handled as a separate change.
Filing it as a follow-up rather than implying this closes the whole class.

### Tests I ran

Added unit tests for the boundary logic (exact fit at the limit, over-by-one,
empty chunk, body already at the limit, and the underflow-safety case), plus a
case asserting `ResponseBodyTooLarge` is non-retryable. Full crate suite,
lints, and formatting all pass on the pinned toolchain (1.95.0):

```
cargo test  -p codex-rmcp-client                                 # 75 passed, 0 failed
cargo clippy -p codex-rmcp-client --all-targets -- -D warnings   # clean
cargo fmt   -p codex-rmcp-client -- --check                      # no diffs
```

### DIFF
--- a/codex-rs/rmcp-client/src/http_client_adapter.rs
+++ b/codex-rs/rmcp-client/src/http_client_adapter.rs
@@ -589,3 +589,41 @@ fn sse_stream_from_body(
     }))
     .boxed()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::within_body_limit;
+
+    #[test]
+    fn allows_chunk_that_exactly_reaches_limit() {
+        assert!(within_body_limit(90, 10, 100));
+    }
+
+    #[test]
+    fn allows_empty_chunk_at_limit() {
+        assert!(within_body_limit(100, 0, 100));
+    }
+
+    #[test]
+    fn rejects_chunk_that_exceeds_limit_by_one() {
+        assert!(!within_body_limit(90, 11, 100));
+    }
+
+    #[test]
+    fn rejects_first_chunk_larger_than_limit() {
+        assert!(!within_body_limit(0, 101, 100));
+    }
+
+    #[test]
+    fn rejects_any_chunk_once_body_is_at_limit() {
+        assert!(!within_body_limit(100, 1, 100));
+    }
+
+    #[test]
+    fn does_not_underflow_when_body_already_exceeds_limit() {
+        // A non-empty chunk is rejected rather than panicking on the internal
+        // subtraction; a zero-length chunk grows nothing and stays allowed.
+        assert!(!within_body_limit(200, 1, 100));
+        assert!(within_body_limit(200, 0, 100));
+    }
+}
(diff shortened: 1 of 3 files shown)


interpreterwork:
Thank you for identifying the unbounded buffered-response path. A response limit is a valid defense-in-depth concern, but this is general inherited MCP transport behavior in `codex-rmcp-client`, not an Open Interpreter-specific integration.

Please propose this to [`openai/codex`](https://github.com/openai/codex), where the maintainers can choose a transport-wide policy and coordinate the buffered JSON/error path with the still-unbounded SSE event path. In particular, the 100 MiB ceiling is a product-level transport decision that should be made upstream rather than carried as a fork-specific divergence.

Closing it here to keep the Open Interpreter delta focused.

PR #1825, opened 2026-07-18
fix(login): store Kimi credentials with owner-only permissions

ethan0905:
### What I found

`save_token` in `login/src/kimi_code.rs` persists the Kimi credential file
(`kimi-code.json`), which contains the **plaintext access and refresh tokens**,
using the default umask:

```rust
fs::create_dir_all(credentials_dir(codex_home))?;   // dir: default perms
...
fs::write(&temp_path, &contents)?;                  // file: default perms
fs::rename(temp_path, path)?;
```

With a typical `022` umask the file lands at `0o644` (world-readable) and the
`credentials/` directory at `0o755`. This is inconsistent with the OpenAI
`auth.json` store in the same crate, which deliberately writes `0o600`
(`login/src/auth/storage.rs:211-214`). `CODEX_HOME` itself is created with a
plain `create_dir_all` (so `0o755` under a default umask), so the parent
directory does not protect these files either.

On a shared or multi-user host — or any setup where `$HOME`/`$CODEX_HOME` is
group- or world-traversable (e.g. the `0o755` home default on Debian/Ubuntu) —
another local user can read `$CODEX_HOME/credentials/kimi-code.json` and lift the
long-lived **refresh token**, which is enough to mint new access tokens and take
over the user's Kimi provider account. Unlike the OpenAI path, these tokens are
neither encrypted nor keyring-backed, so file permissions are their only
protection. (On single-user machines or `0o700`-home distros this isn't
exploitable, so realistically it's a shared-host issue — but a real one there.)

### How I fixed it

Restrict the credential file to `0o600` and the directory to `0o700` on Unix,
mirroring the `auth.json` idiom — and make the guarantee actually hold rather
than best-effort:

- **Directory**: create it `0o700`, and explicitly tighten it if it already
  exists (an earlier version may have left it `0o755`; `DirBuilder` won't change
  an existing directory).
- **File**: remove any stale or pre-planted temp file first, then create it with
  `create_new` + mode `0o600`. This means a leftover temp can't keep its old
  loose mode, a planted **symlink can't be followed** (avoiding a
  truncate-through-symlink write), and a race between the remove and create
  **fails closed** instead of writing through whatever appeared. The mode is also
  set explicitly after open so it's exact regardless of umask.

Non-Unix behavior is unchanged.

### Scope

Focused on the credential file and its directory, which hold the actual secret. I
left the `device_id` file untouched: it's a randomly generated device identifier
sent as a request header, not a credential. Happy to harden it too if maintainers
prefer.

### Tests I ran

Two Unix tests: one asserts a clean `save_token` produces a `0o600` file and
`0o700` directory; the other pre-creates a **world-readable directory and a stale
world-readable temp file** and asserts the result is still `0o600`/`0o700`, i.e.
existing loose permissions are repaired rather than preserved.

```
cargo test  -p codex-login --lib kimi_code                 # 4 passed, 0 failed
cargo clippy -p codex-login --all-targets --no-deps        # clean
cargo fmt   -p codex-login -- --check                      # no diffs
```

(Note: a full `cargo clippy -p codex-login` also lints a pre-existing warning in
the unrelated `codex-core-plugins` dependency, which this change does not touch.)


interpreterwork:
I updated the contributor branch and brought it onto current `main`.

The original owner-only permissions fix is retained, with a few hardening changes:

- existing loose Unix permissions are repaired when credentials are read, not only after the next token refresh/write;
- symlinked credential directories and symlinked credential files on reads fail closed;
- writes still use a fresh owner-only temporary file, replace a file symlink without touching its target, and handle replacement on Windows;
- regression coverage now includes read-time repair and both directory/file symlink cases.

This remains narrowly scoped to Open Interpreter’s Kimi Code login integration. I’m letting the public GitHub checks validate the updated branch now.

PR #1827, opened 2026-07-18
fix(install): detect architecture without RuntimeInformation.OSArchitecture

### LINKED ISSUE #1821: Cannot install on Windows: The property 'OSArchitecture' cannot be found on this object

Could not install on Windows using the provided command:

```
PS C:\Users\kitsune> irm https://www.openinterpreter.com/install.ps1 | iex
iex : The property 'OSArchitecture' cannot be found on this object. Verify that the property exists.
At line:1 char:51
+ irm https://www.openinterpreter.com/install.ps1 | iex
+                                                   ~~~
    + CategoryInfo          : NotSpecified: (:) [Invoke-Expression], PropertyNotFoundException
    + FullyQualifiedErrorId : PropertyNotFoundStrict,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

Image

1. Launch Powershell
2. Enter the provided installation command, `irm https://www.openinterpreter.com/install.ps1 | iex`

_No response_

### Additional information

- Windows 11 Pro 25H2

ethan0905:
Fixes #1821

### What I found

The Windows installer detects the CPU architecture with:

```powershell
$architecture = [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture
```

`RuntimeInformation.OSArchitecture` was added in .NET Framework 4.7.1. On older
Windows PowerShell / .NET Framework hosts it isn't available, so this line throws
**"The property 'OSArchitecture' cannot be found on this object"** — the exact
error reported in #1821 — and the install aborts before doing anything.

### How I fixed it

Try `OSArchitecture` first, and if it isn't available fall back to the
architecture environment variables, which exist on every Windows version:

- `PROCESSOR_ARCHITEW6432` first (holds the real architecture inside a 32-bit
  WOW64 process),
- then `PROCESSOR_ARCHITECTURE`.

`AMD64` maps to `X64` and `ARM64` to `Arm64`, feeding the existing target `switch`
unchanged. A 64-bit OS is already required a few lines above, so no 32-bit case
needs handling.

### Why this fully resolves it

The failure is a single missing-property error, and the fallback is the standard,
universally-available way to read the architecture — so the reported install
failure simply no longer occurs, on both modern and old PowerShell. There's no
behavior change on hosts where `OSArchitecture` already works (that path is tried
first and unchanged).

### Testing

I added `scripts/install/test_install_ps1.ps1` (alongside the existing
`test_install_sh.py`). Rather than duplicate the logic, it **extracts the real
architecture-detection block from `install.ps1`** and runs it, only swapping the
`RuntimeInformation.OSArchitecture` access for a probe the test controls — so it
can simulate a host where that property is missing without needing an actual old
.NET box. It covers:

- **modern hosts** (`OSArchitecture` works) → `x64`/`arm64` resolve unchanged;
- **the #1821 case** (`OSArchitecture` throws) → falls back to
  `PROCESSOR_ARCHITECTURE` (`AMD64`→x64, `ARM64`→arm64);
- **32-bit WOW64 PowerShell** → resolves the real arch from
  `PROCESSOR_ARCHITEW6432`;

plus a parse check that the real `install.ps1` has no syntax errors.

Run it on any PowerShell (Windows PowerShell 5.1 or `pwsh` on macOS/Linux/Windows):

```
pwsh -NoProfile -File scripts/install/test_install_ps1.ps1
```

Output (run on `pwsh` 7 / macOS):

```
PASS: install.ps1 parses cleanly
PASS: modern x64 target -> x86_64-pc-windows-msvc
PASS: modern arm64 target -> aarch64-pc-windows-msvc
PASS: old .NET amd64 target -> x86_64-pc-windows-msvc
PASS: old .NET arm64 target -> aarch64-pc-windows-msvc
PASS: old .NET WOW64 arm64 target -> aarch64-pc-windows-msvc
PASS: old .NET WOW64 amd64 target -> x86_64-pc-windows-msvc

All install.ps1 checks passed.
```

The test is CI-friendly (exits non-zero on failure), so it can be wired into a
`pwsh`-based workflow if desired.

### Testing notes

The fix is defensive on both fronts: a missing-property access that raises a
*terminating* error is caught by `try`/`catch`, and a *non-terminating* one
leaves `$architecture` empty and is caught by the `IsNullOrWhiteSpace` fallback —
so it recovers regardless of how the missing property surfaces on a given host.
The committed test exercises this by making `OSArchitecture` throw against the
real installer block. (Old .NET Framework < 4.7.1 hosts are rare today, so I
validated via that simulation rather than a live legacy box.)

### DIFF
--- a/scripts/install/install.ps1
+++ b/scripts/install/install.ps1
@@ -812,7 +812,30 @@ if (-not [Environment]::Is64BitOperatingSystem) {
     exit 1
 }
 
-$architecture = [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture
+$architecture = $null
+try {
+    $architecture = [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString()
+} catch {
+    # Older Windows PowerShell / .NET Framework (< 4.7.1) does not expose
+    # RuntimeInformation.OSArchitecture, which throws "The property
+    # 'OSArchitecture' cannot be found on this object." Fall back to the
+    # architecture environment variables, which are available everywhere.
+    $architecture = $null
+}
+if ([string]::IsNullOrWhiteSpace($architecture)) {
+    # PROCESSOR_ARCHITEW6432 holds the real architecture inside a 32-bit
+    # (WOW64) process; PROCESSOR_ARCHITECTURE is correct otherwise. A 64-bit
+    # OS is already required above.
+    $envArchitecture = $env:PROCESSOR_ARCHITEW6432
+    if ([string]::IsNullOrWhiteSpace($envArchitecture)) {
+        $envArchitecture = $env:PROCESSOR_ARCHITECTURE
+    }
+    switch ($envArchitecture) {
+        "AMD64" { $architecture = "X64" }
+        "ARM64" { $architecture = "Arm64" }
+        default { $architecture = $envArchitecture }
+    }
+}
 $target = $null
 $platformLabel = $null
 $npmTag = $null
--- a/scripts/install/test_install_ps1.ps1
+++ b/scripts/install/test_install_ps1.ps1
@@ -1,0 +1,105 @@
+#!/usr/bin/env pwsh
+# Tests for scripts/install/install.ps1.
+#
+# These run on any PowerShell (Windows PowerShell 5.1 or `pwsh` on
+# macOS/Linux/Windows):
+#
+#     pwsh -NoProfile -File scripts/install/test_install_ps1.ps1
+#
+# The architecture-detection cases exercise the *real* block extracted from
+# install.ps1 (not a copy), only swapping the RuntimeInformation.OSArchitecture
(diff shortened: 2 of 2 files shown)


interpreterwork:
Thanks for the contribution. This is now superseded by #1831, which has landed on `main` with robust reflected type/property discovery, normalized fallback through `RuntimeInformation`, `PROCESSOR_ARCHITEW6432`, and `PROCESSOR_ARCHITECTURE`, fail-closed handling for unsupported architectures, and public Windows x64/ARM64 coverage under both Windows PowerShell and PowerShell.

Closing this as a duplicate. This PR still helped confirm the installer fallback issue and the need for regression coverage.

ethan0905:
Thanks for taking the time to review my PR. Glad it helped confirm the installer fallback issue and the need for regression coverage. #1831 looks like a more complete solution, happy to see it landed! I'll keep an eye out for other areas where I can help.

PR #1831, opened 2026-07-20
fix(install): fall back when OSArchitecture is missing on Windows

### LINKED ISSUE #1821: Cannot install on Windows: The property 'OSArchitecture' cannot be found on this object

Solaris-star:
## Summary
Windows install (`irm https://www.openinterpreter.com/install.ps1 | iex`) fails on some PowerShell hosts with:

```
The property 'OSArchitecture' cannot be found on this object.
```

The script runs under `Set-StrictMode -Version Latest`, so a missing static property becomes a hard error instead of `$null`.

## Fix
- Probe `RuntimeInformation.OSArchitecture` before reading it
- On failure / absence, map `PROCESSOR_ARCHITECTURE` / `PROCESSOR_ARCHITEW6432` to the same `X64` / `Arm64` labels the switch already expects

## Test plan
- [ ] On Windows PowerShell 5.1 / 7: `irm ... | iex` no longer throws at architecture detection
- [ ] x64 still resolves to `x86_64-pc-windows-msvc`
- [ ] ARM64 still resolves to `aarch64-pc-windows-msvc`

Fixes #1821.


interpreterwork:
I updated this branch as `InterpreterWork` and brought it onto current `main`.

I kept the fallback proposed here, then tightened and tested it:

- RuntimeInformation discovery now uses reflection and safely handles both a missing type and a missing `OSArchitecture` property without directly touching the unavailable static property.
- Architecture resolution has an explicit order: RuntimeInformation, `PROCESSOR_ARCHITEW6432` (the native architecture under WOW64), then `PROCESSOR_ARCHITECTURE`.
- Only ARM64 and x64 are accepted; unsupported 32-bit values fail explicitly.
- A focused PowerShell regression test exercises the fallback and precedence cases without adding a test-only path to the standalone installer.
- `public-ci` now runs that test on Windows x64 and Windows ARM64, under both Windows PowerShell and modern PowerShell.

The extra coverage matters because this installer is a standalone release and auto-update boundary. I am waiting for the new Windows checks before merge.

PR #1834, opened 2026-07-21
fix(apply-patch): fall back from EOF when seek_sequence misses end match

### LINKED ISSUE #1830: apply-patch: seek_sequence(eof=true) never falls back to scanning from start (documented but unimplemented) → false "Failed to find expected lines"
## Summary

`codex-rs/apply-patch/src/seek_sequence.rs` `seek_sequence()` has a documented but **unimplemented** behavior: when `eof` is `true`, it is supposed to try the end-of-file position first and then **fall back to searching from `start`**. In practice the `eof` branch restricts *every* match pass (exact, rstrip, trim, unicode-normalise) to a **single** position — the EOF position — and never falls back. As a result, when a patch chunk is marked `*** End of File` but its context/old lines do not correspond to the actual last lines of the file, `apply_patch` fails with `Failed to find expected lines` even though the lines exist elsewhere in the file and could be located by the existing fuzzy machinery.

## Root cause

`codex-rs/apply-patch/src/seek_sequence.rs`:

```rust
let search_start = if eof && lines.len() >= pattern.len() {
    lines.len() - pattern.len()            // == the single EOF position
} else {
    start
};
// Exact match first.
for i in search_start..=lines.len().saturating_sub(pattern.len()) {
    ...
}
// (then rstrip / trim / normalise passes, all using the same `search_start..=` range)
```

When `eof == true`, `search_start == lines.len() - pattern.len()`, which equals the loop's upper bound, so each pass iterates exactly **one** position (the EOF position). There is no second pass scanning from `start`.

This contradicts the function's own doc comment:

> When `eof` is true, we first try starting at the end-of-file (so that patterns intended to match file endings are applied at the end), **and fall back to searching from `start` if needed.**

The fact that four increasingly-lenient match passes (including full Unicode punctuation normalisation) exist, only to be restricted to a single position when `eof` is set, is a strong signal that the fallback was intended and its absence is a bug.

## How it is reached

`compute_replacements()` in `codex-rs/apply-patch/src/lib.rs` calls:

```rust
seek_sequence::seek_sequence(original_lines, pattern, line_index, chunk.is_end_of_file);
```

`chunk.is_end_of_file` is set to `true` by the parser whenever a chunk is followed by an `*** End of File` marker. For chunks with non-empty `old_lines` (context/removed lines), the `old_lines.is_empty()` early-return is skipped and `seek_sequence` is invoked with `eof = true`. (The pure-addition case `old_lines.is_empty()` returns early and never reaches `seek_sequence`.)

So the failure occurs when a chunk has **both** non-empty `old_lines` **and** `is_end_of_file == true`, and those `old_lines` are not the file's actual last lines. Realistic triggers:

- The model's view of the file is stale / truncated, so it marks a region as end-of-file that isn't.
- The file on disk grew after the diff was produced.
- The file has a trailing blank line (a single trailing `""` survives the `split('\n')` + `pop()` in `lib.rs`) so the real last non-blank line is not at the EOF position.

In all of these, the lines the patch wants to anchor on *do* exist earlier in the file, but `apply_patch` reports `Failed to find expected lines` and aborts the whole patch.

## Minimal repro

`seek_sequence` is a pure `pub(crate)` function with existing unit tests. This assertion currently **fails** (returns `None`), even though the pattern exists at index 0:

```rust
let lines   = vec!["a".to_string(), "b".to_string(), "c".to_string(), "d".to_string()];
let pattern = vec!["a".to_string(), "b".to_string()];
// eof=true forces a search restricted to the single EOF position (index 2),
// where "c"/"d" sit, so no match is found even though "a"/"b" is at index 0.
assert_eq!(seek_sequence(&lines, &pattern, 0, true), None); // bug
// expected with documented fallback: Some(0)
```

## Suggested fix

Build the list of candidate start positions once — EOF position first when `eof` is set, then the full `start..=last` range — and iterate that list in each strictness pass. This honours the documented contract ("try EOF first, fall back to start"), preserves EOF preference, and is strictly more lenient (any case that currently matches at EOF still matches at EOF first):

```rust
let last = lines.len().saturating_sub(pattern.len());
let mut candidates: Vec<usize> = Vec::with_capacity(last.saturating_sub(start) + 2);
if eof {
    candidates.push(last);
}
candidates.extend(start..=last);

// then each pass:
for &i in &candidates { /* unchanged per-position check */ }
```

Verified standalone (rustc 1.94, `--test`): with this change the repro above returns `Some(0)`, the EOF-preference case (pattern at both start and EOF) still returns the EOF index, and all existing tests in the module still pass. `cargo clippy` with the crate's deny-level lints (`needless_collect`, `redundant_clone`, …) is clean (the `Vec` is consumed by four passes and needs a specific ordering, so it is not a single-use collect).

## Impact

Correctness/robustness of patch application — `apply_patch` is the mechanism behind every model-driven file edit. Currently a stale/inaccurate `*** End of File` marker on a chunk with context lines fails the entire patch instead of degrading gracefully to the existing fuzzy locator.

Solaris-star:
## Summary
- `seek_sequence(..., eof=true)` is documented to try the end-of-file position first, then fall back to scanning from `start`.
- The old implementation locked every match pass to a single EOF index, so `*** End of File` chunks with non-trailing context failed with `Failed to find expected lines`.
- Extract `seek_from` and implement EOF-then-fallback; keep the existing exact / rstrip / trim / Unicode-normalise ladder.

Fixes #1830

## Test plan
- [x] `cargo test -p codex-apply-patch --lib seek_sequence` (7 passed)
- [x] New coverage: prefer end match, fall back from start, no-match returns `None`
- [ ] Existing apply-patch suite still green in CI

### DIFF
--- a/codex-rs/apply-patch/src/seek_sequence.rs
+++ b/codex-rs/apply-patch/src/seek_sequence.rs
@@ -26,11 +26,24 @@ pub(crate) fn seek_sequence(
     if pattern.len() > lines.len() {
         return None;
     }
-    let search_start = if eof && lines.len() >= pattern.len() {
-        lines.len() - pattern.len()
-    } else {
-        start
-    };
+
+    // Prefer an end-of-file match when requested, then fall back to a normal
+    // scan from `start` so patterns that only appear earlier still resolve.
+    if eof {
+        let eof_start = lines.len() - pattern.len();
+        if let Some(idx) = seek_from(lines, pattern, eof_start) {
+            return Some(idx);
+        }
+        if start < eof_start {
+            return seek_from(lines, pattern, start);
+        }
+        return None;
+    }
+
+    seek_from(lines, pattern, start)
+}
+
+fn seek_from(lines: &[String], pattern: &[String], search_start: usize) -> Option<usize> {
     // Exact match first.
     for i in search_start..=lines.len().saturating_sub(pattern.len()) {
         if lines[i..i + pattern.len()] == *pattern {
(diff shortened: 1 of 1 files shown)


interpreterwork:
Thanks for the contribution. The change is technically sensible, but it modifies general Codex apply-patch behavior rather than an Open Interpreter-specific delta.

This project focuses on the additional harnesses and product integration layered on top of upstream Codex, so we want general runtime fixes to land in `openai/codex` first rather than creating a separate behavior here.

Please consider opening this as a PR against https://github.com/openai/codex. If it is merged upstream, Open Interpreter will pick it up through our stable Codex release sync because this project is based on upstream.

I’m closing this PR here to keep the Open Interpreter delta focused. Thanks again.

PR #1845, opened 2026-07-27
Fix full cloud CI after Codex 0.145 sync

interpreterwork:
Restores the Open Interpreter project-config boundary, fixes post-sync test expectations and Bazel metadata, removes an unused workspace dependency, and limits the Docker remote-executor lane to tests that explicitly opt into that environment.\n\nValidation is intentionally running in GitHub Actions before merge.


interpreterwork:
Superseded after a finite-delta audit. The replacement drops all edits to upstream-owned tests and restores the upstream remote-test selection.

PR #1855, opened 2026-07-29
feat: support pre-release (beta/canary) versions in Rust release workflows

### LINKED ISSUE #1854: Support pre-release (beta/canary) versions in Rust release workflows
### Description
Currently, the Rust release workflows (`rust-release.yml` and `rust-release-publish-existing.yml`) do not support publishing pre-release versions (such as beta, canary, or release candidates). 

Specifically:
1. The `push` trigger in `rust-release.yml` is restricted to `rust-v*.*.*`, which does not match pre-release tags like `rust-v0.0.17-beta.0`.
2. The tag validation regex in `rust-release.yml` enforces a strict semver without any pre-release suffix: `^rust-v[0-9]+\.[0-9]+\.[0-9]+$`.
3. The `prerelease` parameter of the `softprops/action-gh-release` action is hardcoded to `false` in both workflows, so even if a pre-release tag were triggered, it would be published as a full stable release.

### Proposal
Modify the release workflows to:
- Expand the tag triggers to include `rust-v*.*.*-*` to capture pre-release tags.
- Update the tag validation regex to `^rust-v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.-]+)?$` to allow pre-release suffixes.
- Dynamically determine if the version is a pre-release (by checking if it contains a hyphen) and set the `prerelease` input of `softprops/action-gh-release` accordingly.

ImBIOS:
### Description
Currently, the Rust release workflows (`rust-release.yml` and `rust-release-publish-existing.yml`) do not support publishing pre-release versions (such as beta, canary, or release candidates).

This PR modifies the release workflows to:
- Expand the tag triggers to include `rust-v*.*.*-*` to capture pre-release tags.
- Update the tag validation regex to `^rust-v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.-]+)?$` to allow pre-release suffixes.
- Dynamically determine if the version is a pre-release (by checking if it contains a hyphen) and set the `prerelease` input of `softprops/action-gh-release` accordingly.

This allows releasing beta/canary versions of the Open Interpreter CLI.

Closes #1854

### DIFF
--- a/.github/workflows/rust-release-publish-existing.yml
+++ b/.github/workflows/rust-release-publish-existing.yml
@@ -74,13 +74,23 @@ jobs:
           sha256sum open-interpreter-package-*.tar.gz open-interpreter-package-*.tar.zst \
             > codex-package_SHA256SUMS
 
+      - name: Determine prerelease status
+        id: prerelease
+        shell: bash
+        run: |
+          if [[ "${{ inputs.version }}" == *-* ]]; then
+            echo "prerelease=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "prerelease=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Publish release
         uses: softprops/action-gh-release@72f2c25fcb47643c292f7107632f7a47c1df5cd8 # v2.5.0
         with:
           tag_name: ${{ inputs.tag }}
           name: Open Interpreter ${{ inputs.version }}
           draft: false
-          prerelease: false
+          prerelease: ${{ steps.prerelease.outputs.prerelease }}
           files: |
             dist/open-interpreter-package-*.tar.gz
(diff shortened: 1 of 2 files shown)


interpreterwork:
Thanks for the proposal. We are not adopting this prerelease workflow as written. The existing rust-v*.*.* Actions glob already matches suffixed tags; the proposed regex admits malformed semver; manual dispatch remains insufficiently validated; and the public installers currently recognize only alpha and beta suffixes, so publishing canary or rc tags would create release paths that the supported installers cannot consume consistently. Prerelease publication needs an explicit release-channel decision plus end-to-end installer and update validation across macOS, Linux, and Windows. Closing this until that complete design is approved.

PR #1857, opened 2026-07-29
feat: Add support for Gemini 3.6 Flash and Gemini 3.5 Flash-Lite

### LINKED ISSUE #1856: feat: Add support for Gemini 3.6 Flash and Gemini 3.5 Flash-Lite models
Google recently launched the new Gemini models, including **Gemini 3.6 Flash** and **Gemini 3.5 Flash-Lite**.

ImBIOS:
Closes #1856

### Description
This PR updates the model compatibility catalog by running the `write_model_compatibility_catalog.py` script. This adds support for Google's newly launched Gemini models:
- **Gemini 3.6 Flash** (`gemini-3.6-flash`)
- **Gemini 3.5 Flash-Lite** (`gemini-3.5-flash-lite`)

These models are now listed in the compatibility catalog with their full features (such as reasoning effort, parallel tool calling, search tools, etc.) to enable users to use them seamlessly.

### Testing
- Ran `just test -p codex-api` successfully (all 173 tests passed).

### DIFF
--- a/codex-rs/codex-api/model_compatibility_catalog.json
+++ b/codex-rs/codex-api/model_compatibility_catalog.json
@@ -528,6 +528,25 @@
       "context_window": null
     },
     {
+      "id": "aiml/openai/gpt-image-2",
+      "aliases": [
+        "openai/gpt-image-2"
+      ],
+      "description": null,
+      "visibility": "hide",
+      "supported_parameters": [],
+      "supported_reasoning_levels": [],
+      "supports_thinking_toggle": false,
+      "reasoning_control": "none",
+      "supports_parallel_tool_calls": false,
+      "supports_search_tool": false,
+      "input_modalities": [
+        "text",
+        "image"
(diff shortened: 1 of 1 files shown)


interpreterwork (changes requested):
The generated compatibility catalog refresh is valid: I reproduced the generator output exactly, and both Gemini model IDs are already present in the generated provider catalog. This PR currently also includes the unrelated build-interpreter-release.sh commit, which is separately proposed in #1858. Please rebase onto current main and drop that commit so this PR contains only the generated compatibility catalog refresh. After rebasing, rerun the live-source generator and update the test plan with the exact generator command and the focused compatibility/model checks.

interpreterwork (approved):
Maintainer edit applied: the unrelated release-script commit has been removed. The PR now contains only the generated compatibility catalog refresh, which I reproduced from the live-source generator.

PR #1858, opened 2026-07-29
feat(scripts): enhance build-interpreter-release.sh with option parsing and path resolution

ImBIOS:
### Summary

Improves `scripts/build-interpreter-release.sh` to enhance usability, safety, and local build speed:

1. **Flexible CLI Options & Equals Syntax:**
   - Added support for `--cargo-profile` / `--profile` (defaults to `release`).
   - Added `-j` / `--jobs` for controlling Cargo parallel build jobs.
   - Added support for `--key=value` option parsing across all CLI flags (`--target`, `--install-dir`, `--home`, `--profile`, `--jobs`).

2. **Parallel Build Performance:**
   - Removed the default `CARGO_BUILD_JOBS=1` fallback which previously forced single-threaded compilation. Cargo now defaults to auto-detecting all CPU cores unless `-j` or `CARGO_BUILD_JOBS` is explicitly specified.

3. **Absolute Path Resolution:**
   - Canonicalizes `install_dir` and `interpreter_home` to absolute paths before running python package build tools or creating symlinks, fixing broken links when passing relative paths (e.g. `--home ./local-home`).

4. **Safety & Robustness:**
   - Ensured `mkdir -p` is called for link target directories in `replace_symlink()`.
   - Hardened `cleanup()` to verify `$staging_dir` is non-empty before running `rm -rf`.
   - Added an executable check before verifying the binary with `--version`.

### DIFF
--- a/scripts/build-interpreter-release.sh
+++ b/scripts/build-interpreter-release.sh
@@ -9,8 +9,7 @@ Options:
   --target <target>        Rust target triple (defaults to host platform)
   --install-dir <dir>      Visible bin directory for shims
   --home <dir>             Open Interpreter home directory
-  --cargo-profile <prof>   Cargo profile to build (default: release)
-  -j, --jobs <N>           Number of parallel Cargo build jobs (default: auto)
+  -j, --jobs <N>           Number of parallel Cargo build jobs (default: 1)
   -h, --help               Show this help message
 
 Builds a local Open Interpreter standalone package using the same package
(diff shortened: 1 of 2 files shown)


interpreterwork (changes requested):
Thanks for working on the build script. This needs revision before it can land. The canonical build-interpreter-release.sh entrypoint must continue to produce a release artifact, so it should not accept an arbitrary Cargo profile. Please also preserve the intentional single-job default unless a documented requirement justifies changing the low-resource build behavior. Split path canonicalization and symlink safety from build-performance behavior, and add focused tests for argument parsing, relative paths, missing values, and symlink refusal. A linked issue describing the intended behavior and tradeoffs would help keep the change scoped.

interpreterwork (approved):
Maintainer edit applied: the canonical script remains release-only, preserves the one-job default, validates job counts, and retains the scoped option/path safety improvements. Added five end-to-end script tests; all pass locally.

PR #1863, opened 2026-08-04
feat(core): set AI_AGENT for shell commands

complynx:
## Summary

- expose `AI_AGENT=open-interpreter` to shell commands launched by Open Interpreter
- preserve an explicit non-blank marker supplied by a wrapper or orchestrator
- handle case-insensitive Windows environment keys without leaving duplicates
- add focused coverage for the default and preservation behavior

## Why a universal marker

Tools launched by coding agents often need to adapt telemetry, UX, logging, or safety behavior based on whether they are running under an agent. Today they generally have to maintain agent-specific process-tree and environment heuristics. A shared `AI_AGENT=<slug>` convention lets the producer that already knows its identity declare it once, avoiding an N-by-M matrix of consumer heuristics as agents and tools multiply.

This is deliberately an advisory marker, not a security boundary. Open Interpreter only supplies its own slug when the variable is missing or blank, so wrappers and higher-level orchestrators retain control of any explicit non-blank value.

## Validation

- `git diff --check`
- attempted `bazel test --test_output=errors --test_filter=exec_env //codex-rs/core:core-unit-tests`; repository analysis stops before compilation because `codex_rust_crate()` rejects the existing `binary_test_target_compatible_with` argument in `codex-rs/windows-sandbox-rs/BUILD.bazel`
- attempted the repository formatter; this checkout environment does not provide its required `just`, Cargo, or dotslash tools

### DIFF
--- a/codex-rs/core/src/exec_env.rs
+++ b/codex-rs/core/src/exec_env.rs
@@ -8,6 +8,9 @@ use std::collections::HashMap;
 
 pub use codex_protocol::shell_environment::CODEX_THREAD_ID_ENV_VAR;
 
+const AI_AGENT_ENV_VAR: &str = "AI_AGENT";
+const OPEN_INTERPRETER_AI_AGENT: &str = "open-interpreter";
+
 /// Informational name of the active permission profile. Child processes can
 /// overwrite this value, so it must not be treated as proof of enforcement.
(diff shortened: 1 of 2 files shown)


interpreterwork:
This changes generic Codex shell-command behavior rather than an Open Interpreter-specific integration. Please redirect it to openai/codex; an accepted upstream change will flow into Open Interpreter. Closing here.

PR #1880, opened 2026-08-19
perf: reduce repeated root-depth checks in file search

brandonbirch1011:
## Summary

This small optimization avoids recomputing the root depth for each matching search directory while resolving a file path against multiple roots.

## Why

get_file_path() was repeatedly calling components().count() on the current best root to decide whether a newer candidate should replace it, which adds unnecessary work on every path match during file searches.

## Changes

- Keep the best root depth alongside the best match and only compare against the stored depth.
- Preserve the existing behavior of preferring the shallowest matching root and keeping the earliest match on ties.
- Added a regression test covering nested search roots.

## Validation

I attempted to run the targeted Rust test suite, but the local environment is missing the Visual Studio C++ linker (link.exe), so compilation cannot complete in this container.

### DIFF
--- a/codex-rs/file-search/src/lib.rs
+++ b/codex-rs/file-search/src/lib.rs
@@ -378,21 +378,24 @@ fn build_override_matcher(
 }
 
 fn get_file_path<'a>(path: &'a Path, search_directories: &[PathBuf]) -> Option<(usize, &'a str)> {
-    let mut best_match: Option<(usize, &Path)> = None;
+    let mut best_match: Option<(usize, usize, &'a Path)> = None;
     for (idx, root) in search_directories.iter().enumerate() {
-        if let Ok(rel_path) = path.strip_prefix(root) {
-            let root_depth = root.components().count();
-            match best_match {
-                Some((best_idx, _))
-                    if search_directories[best_idx].components().count() >= root_depth => {}
-                _ => {
-                    best_match = Some((idx, rel_path));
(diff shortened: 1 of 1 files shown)


KillianLucas (changes requested):
Thanks — this changes an intentional root-selection invariant. Project instructions use the nearest (deepest) matching ancestor as the project root. Selecting a higher outer match can cross a nested-project boundary. Please preserve nearest-root semantics while caching the depth, and add a nested-marker regression proving the inner marker wins; then we can re-review.

KillianLucas:
Thanks for working on this. I compared the change with the current OIX tree and upstream Codex. As written, the comparator is reversed: the existing best_depth >= root_depth guard retains the deepest (nearest) matching root, while best_depth <= root_depth selects the outermost root. For nested roots, the new test would therefore assert the wrong root index and relative path.

This helper is generic Codex multi-root file-search behavior, and upstream currently carries the same deepest-root implementation, introduced in openai/codex commit d59685f6d4f. It is not part of OIX’s documented finite delta, so we should avoid maintaining a downstream fork here.

Please take the optimization upstream to openai/codex. If upstream wants the refactor, it should preserve the deepest-root result, carry (idx, depth, rel_path), and cover equal-depth tie behavior as well. I’m closing this downstream PR to keep the fork aligned; an accepted upstream version will arrive through the normal sync.

PR #1882, opened 2026-08-20
Release Open Interpreter 0.0.40

interpreterwork:
## Open Interpreter boundary

- [x] I read `AGENTS.md` and `README.md`.
- [x] This is release metadata for Open Interpreter, not a change to general Codex behavior.
- [x] I considered whether this belongs in `openai/codex`; it does not.
- [x] This change is specific to Open Interpreter standalone packaging and update discovery.

## Why this belongs in Open Interpreter

This prepares the Open Interpreter release version consumed by its standalone packages, updater, and TUI update notices after the stable Codex `rust-v0.148.0` integration.

## Summary

- bump the Rust workspace from 0.0.39 to 0.0.40
- update the lockfile and version-bearing TUI snapshots
- keep the release delta limited to Open Interpreter release metadata

## Test plan

- `cargo fmt --check`
- `just fmt`
- locked Cargo metadata resolves the workspace as 0.0.40
- no remaining 0.0.39 references
- `git diff --check`
- full public Linux and Windows CI

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -460,7 +460,7 @@ checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "app_test_support"
-version = "0.0.39"
+version = "0.0.40"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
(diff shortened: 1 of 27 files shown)


swithinchan (commented):
Dear Killian

Thanks for bringing in the wonderful package of open interpreter, i hope to reflect a recent bug just upgraded 0.0.40, the response using deepseek is always the following:
■ unsupported operation: non-Responses wire APIs require the compatibility
transport

I have already consulted the agent on your website, change the config of harness to "" like below, no default, but it always return the same thing. it never happens before this upgrade. 

model = "deepseek-v4-pro"
harness = "" # optional – explicitly clear it

PR #1899, opened 2026-09-10
fix(kimi-code): reject non-public FetchURL targets

cestercian:
## Summary
FetchURL (kimi-code harness) advertised a public-only contract but performed a raw `reqwest` GET after only an `http`/`https` scheme check. Private/loopback/link-local destinations and unsafe redirect hops were reachable in-process, bypassing the sandbox network-proxy SSRF policy.

## Changes
- Reuse `codex_network_proxy::is_non_public_ip` for IP literals (loopback, RFC1918, link-local including `169.254.169.254`, CGNAT, etc.)
- Reject `localhost` / `*.localhost` before connect
- DNS lookup with fail-closed behavior; reject if any resolved address is non-public
- Disable automatic redirects; revalidate each hop (max 10)
- Cap response body at 1 MiB
- Defensive unit tests with crafted URLs / injected DNS stubs (no live third-party abuse)

## Test plan
- [x] `just test -p codex-core fetch_url` (rejects private/loopback/metadata-style/unsafe schemes; allows public)
- [ ] CI on this PR

Reported privately to `killian@openinterpreter.com` per SECURITY.md.


interpreterwork (changes requested):
Thanks for tackling this. The overall direction is right, but two security properties need to be fixed before merge:

1. `ensure_public_http_url` resolves and validates the hostname, then `reqwest` performs a separate resolution when connecting. A rebinding or time-of-check/time-of-use change can therefore pass validation with a public address and connect to a non-public address. Please bind the validated resolution to the actual connection, or route the request through the existing network-policy enforcement so validation and connection cannot diverge. Add a regression that proves a second DNS answer cannot redirect the connection privately.

2. The response-size check does not enforce a 1 MiB transfer limit when `Content-Length` is absent or false: `.text().await` buffers the complete response before `truncate`. Please stream the body and stop reading once the bounded limit is reached, with coverage for an oversized chunked response.

Once those are addressed, this should be in good shape for focused testing and CI.

interpreterwork (approved):
The updated implementation binds validated DNS results to the connection, revalidates every redirect hop, and enforces the response limit while streaming. The focused regressions cover the two requested security properties. Approved pending CI.

